PR #35, opened 2026-05-08
feat(reading): preference toggle endpoint + HTMX style swap (#16)

### LINKED ISSUE #16: READ-2 — Preference toggle endpoint + HTMX <style> swap

cubrox:
## Ticket

Closes #16 — READ-2: Preference toggle endpoint + HTMX `<style>` swap.

## Summary

The user can now **actually change their reading preferences** and see the change apply instantly. Each click on a sidebar button posts to `/preferences/{key}`, upserts the value, and returns a fresh `<style id='reading-surface-style'>` fragment — HTMX swaps it in place via `outerHTML`. No page reload, no flicker.

This closes the loop on the Reading Surface epic's first arc: combined with READ-1, the user gets a configurable reading view they can tune to their preference and have the preference persist across sessions.

## Changes Made

- **CREATE** [app/services/reading/options.py](app/services/reading/options.py) — `PREFERENCE_OPTIONS` allow-list (per-key valid values), `coerce_value()` (form-string → typed value), `label_for()` (UI button text).
- **CREATE** [app/services/reading/preferences.py](app/services/reading/preferences.py) — `upsert_preference()` with read-merge-write semantics + unchanged-value short-circuit.
- **MODIFY** [app/api/reading.py](app/api/reading.py) — new `POST /preferences/{key}` route. Validates key + value against allow-lists, upserts, returns `<style>` fragment. The GET route now also passes `preference_options` and `label_for` to the page context for sidebar rendering.
- **CREATE** [templates/fragments/reader_style.html](templates/fragments/reader_style.html) — extracted `<style>` block, used by both the page (via `{% include %}`) and the toggle endpoint (via `TemplateResponse`).
- **MODIFY** [templates/pages/reading.html](templates/pages/reading.html) — uses the fragment include + adds the `<details>` sidebar with HTMX-attributed buttons for each preference.
- **MODIFY** [static/style.css](static/style.css) — sidebar styling: `.reader-controls` layout + `aria-pressed` outline marker for active options.
- **CREATE** [tests/test_preferences.py](tests/test_preferences.py) — 17 tests covering each DoD assertion + adversarial CSS-injection attempt.

## Design notes

- **Allow-list validation is the load-bearing safety check.** Both the key (one of seven hardcoded names) AND the value (one of a per-key allow-list) must be valid before any DB write or template render. A value outside the list returns 422. Pinned by `test_value_outside_allow_list_returns_422` AND `test_css_injection_attempt_returns_422` (which posts `red;}body{display:none;}` to /preferences/bg).
- **Read-merge-write, not blind overwrite.** The upsert function reads the existing row's `values` dict, merges in `{key: value}`, writes back. Setting `size` then `line_height` produces a row with BOTH keys set. Pinned by `test_second_toggle_merges_into_existing_values`.
- **Short-circuit on unchanged value.** Clicking the same button twice produces only one DB UPDATE. Pinned by `test_unchanged_value_does_not_update_updated_at` which asserts `updated_at` stays the same on the second post.
- **`bionic_enabled` coerces 'true'/'false' form strings to actual booleans.** Form data arrives as strings; the boolean key needs explicit coercion before the allow-list check (which contains `True`, `False` not `"true"`, `"false"`).
- **Sidebar accessibility from the start**: semantic `<fieldset>` + `<legend>` + `<button type="button">` with `aria-pressed`. The active option for each key is outlined via CSS keyed off `[aria-pressed="true"]`. Keyboard users can tab through; screen readers announce both the legend and the pressed state.
- **Fragment template extraction** lets the page include the same `<style>` block on first paint that the POST endpoint returns on toggle. Single source of truth for the variable layout.

## Reviewer-flagged dependent invariant honored

The READ-1 reviewer flagged that the `| safe` filter on CSS values in the `<style>` block requires WRITE-side allow-list validation to remain safe. This PR is that validation. The `test_css_injection_attempt_returns_422` test pins it.

- Bionic-emphasis text transform (the visual effect when `bionic_enabled=true` is set). **READ-3 (#17)**.
- HTMX OOB swaps to also update aria-pressed on sibling buttons after a toggle. Currently aria-pressed only reflects page-load state; live updates would need the route to return the sidebar fragment too. Polish for a future ticket.
- AUTH-3's cookie-reissue bug fix. The /preferences endpoints return their own HTMLResponse so they're affected by the same FastAPI Response.set_cookie issue documented yesterday. Cookies stay valid for full 30 days regardless; only the rolling-window extension is no-op. Same scope as a future middleware fix.

### Automated tests

- [x] 17 new tests pass locally
- [x] Full suite still green (119 / 119)
- [x] Coverage 98.65% overall; new modules at 100%
- [x] Pre-push hook passed

```bash
# 1. Sign in
# 2. Paste a passage (/passages/new)
# 3. Land on /read/<uuid>
# 4. Open "Reading preferences" details
# 5. Click 24px size button — text resizes WITHOUT page reload
# 6. Click 1.8 line-height button — spacing changes, size stays 24px
# 7. Refresh page — preferences persist
# 8. Try inspecting + editing a hx-vals attribute to send a forged
#    value — see 422 in network tab
```

## Checklist

- [x] All tests pass (119/119, 98.65% coverage)
- [x] Code follows project conventions
- [x] No lint/type errors
- [x] Reviewer's dependent invariant from READ-1 honored
- [ ] Human merge

### DIFF
--- a/app/api/reading.py
+++ b/app/api/reading.py
@@ -1,22 +1,23 @@
-"""Reading-view route.
+"""Reading-view + preference-toggle routes.
 
-GET /read/{passage_id} renders the configurable reading surface for a
-single passage. The user's preferences are inlined into the template's
-`<style>` block as CSS variables — first paint already shows the
-correctly-styled text, no client-side reflow.
+GET  /read/{passage_id}     — render the configurable reading surface
+POST /preferences/{key}     — set one preference, return the swappable
+                              <style> fragment for HTMX outerHTML
 
-Owner check: a user can only view their own passages. Other users'
-passages return 404 (not 403) — same response shape as a nonexistent
-UUID, so the existence of any specific passage isn't leaked.
+Owner check on the GET: a user can only view their own passages. Other
+users' passages return 404 (not 403) — same response shape as a
+nonexistent UUID, so the existence of any specific passage isn't leaked.
 
 Per ADR-005 (HTMX, no SPA), all reading-state lives in the URL + cookie
-+ DB; no client-side state container.
++ DB; no client-side state container. The POST returns ONLY the
+`<style id="reading-surface-style">` fragment so HTMX can swap it
+in place without touching anything else on the page.
 """
 
 import uuid
 from typing import Annotated
 
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, Form, HTTPException, Request
 from fastapi.responses import HTMLResponse
 from sqlmodel import Session, select
 
@@ -26,6 +27,12 @@ from app.models.preference import Preference
 from app.models.user import User
 from app.services.identity.session import current_user
 from app.services.reading.defaults import with_defaults
+from app.services.reading.options import (
+    PREFERENCE_OPTIONS,
+    coerce_value,
+    label_for,
+)
+from app.services.reading.preferences import upsert_preference
 from app.templates import templates
 
 router = APIRouter()
@@ -45,7 +52,8 @@ def read_passage(
 
     Loads the passage (filtered by ownership), loads the user's stored
     preferences (or falls back to defaults if no row exists), renders
-    the full HTML page with the CSS-variable block already populated.
+    the full HTML page with the CSS-variable block already populated
+    AND the preference-toggle sidebar wired up.
     """
     passage = session.exec(
         select(Passage).where(Passage.id == passage_id, Passage.user_id == user.id)  # type: ignore[arg-type]
@@ -62,5 +70,54 @@ def read_passage(
     return templates.TemplateResponse(
         request=request,
         name="pages/reading.html",
-        context={"passage": passage, "prefs": prefs},
+        context={
+            "passage": passage,
+            "prefs": prefs,
+            "preference_options": PREFERENCE_OPTIONS,
+            "label_for": label_for,
+        },
+    )
+
+
+@router.post("/preferences/{key}", response_class=HTMLResponse)
+def update_preference(
+    request: Request,
+    key: str,
+    user: CurrentUser,
+    session: SessionDep,
+    value: Annotated[str, Form()],
+) -> HTMLResponse:
+    """Set ONE preference and return the swappable <style> fragment.
+
+    Returns a fragment, not a full page — HTMX `outerHTML` swaps it
+    into `#reading-surface-style` without touching anything else.
+
+    Validation gates user input against PREFERENCE_OPTIONS in
+    app/services/reading/options.py — both the key and the value must
+    be allow-listed. This is the dependent invariant the READ-1
+    reviewer flagged: without it, the `| safe` filter in the template
+    would let a user inject arbitrary CSS.
+    """
+    if key not in PREFERENCE_OPTIONS:
+        raise HTTPException(status_code=422, detail=f"Unknown preference key: {key!r}")
+
+    coerced = coerce_value(key, value)
+    if coerced is None or coerced not in PREFERENCE_OPTIONS[key]:
+        raise HTTPException(
+            status_code=422,
+            detail=f"Invalid value for {key!r}",
+        )
+
+    upsert_preference(user_id=user.id, key=key, value=coerced, session=session)
+    session.commit()
+
+    # Re-read for the freshest merged view (handles both the new-row
+    # case and the existing-row case uniformly).
+    stored_pref = session.get(Preference, user.id)
+    prefs = with_defaults(stored_pref.values if stored_pref else None)
+
+    return templates.TemplateResponse(
+        request=request,
+        name="fragments/reader_style.html",
+        context={"prefs": prefs},
     )
--- a/app/api/reading.py
+++ b/app/api/reading.py
@@ -31,6 +31,7 @@ from app.services.reading.options import (
     PREFERENCE_OPTIONS,
     coerce_value,
     label_for,
+    value_for_form,
 )
 from app.services.reading.preferences import upsert_preference
 from app.templates import templates
@@ -75,6 +76,7 @@ def read_passage(
             "prefs": prefs,
             "preference_options": PREFERENCE_OPTIONS,
             "label_for": label_for,
+            "value_for_form": value_for_form,
         },
     )
 
--- a/app/services/reading/options.py
+++ b/app/services/reading/options.py
@@ -75,3 +75,19 @@ def label_for(key: str, value: Any) -> str:
     sidebar template to render button text.
     """
     return PREFERENCE_LABELS.get((key, value), str(value))
+
+
+def value_for_form(value: Any) -> str:
+    """Render a preference value as the string the route expects to receive.
+
+    The route's `coerce_value()` accepts lowercase 'true'/'false' for the
+    bionic_enabled boolean. Python's str(True) is 'True' (capitalized),
+    so a naive template render of `{{ opt | string }}` would emit a value
+    the route then rejects with 422. This helper produces the lowercase
+    form for booleans and str(value) for everything else.
+    """
+    if value is True:
+        return "true"
+    if value is False:
+        return "false"
+    return str(value)
--- a/app/services/reading/options.py
+++ b/app/services/reading/options.py
@@ -1,0 +1,77 @@
+"""Allow-list of valid preference keys and per-key valid values.
+
+This is the **dependent invariant** the READ-1 reviewer flagged: any
+user-supplied preference value MUST be validated against this allow-list
+before it reaches the rendered `<style>` block. Without it, a user who
+posted `value=red;}body{display:none}` would get that string injected
+verbatim into CSS via the `| safe` filter on the template (which is
+necessary because legitimate CSS values like font stacks contain
+apostrophes that auto-escape would mangle).
+
+The keys here are the SAME keys used in app/services/reading/defaults.py.
+Values are CSS-ready strings (or booleans for toggles) — directly
+substitutable into `var(--reader-*)` declarations.
+"""
+
+from typing import Any
+
+# Per-key allow-lists. Each list's first entry is the default (matches
+# app/services/reading/defaults.py).
+PREFERENCE_OPTIONS: dict[str, list[Any]] = {
+    "font": [
+        "Georgia, 'Iowan Old Style', 'Charter', 'Bitstream Charter', serif",
+        "system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', sans-serif",
+        "ui-monospace, 'Cascadia Mono', 'Source Code Pro', Menlo, monospace",
+    ],
+    "size": ["14px", "16px", "18px", "20px", "24px", "28px"],
+    "line_height": ["1.4", "1.5", "1.6", "1.8", "2.0"],
+    "bg": ["#ffffff", "#f5e6d3", "#1a1a1a"],  # white, sepia, dark
+    "fg": ["#1a1a1a", "#3d2914", "#e8e8e8"],  # near-black, brown, light
+    "max_width": ["55ch", "65ch", "75ch", "85ch"],
+    "bionic_enabled": [True, False],
+}
+
+# Friendly labels for sidebar UI buttons. Keyed by (preference_key, value).
+# Falls back to str(value) when no entry exists.
+PREFERENCE_LABELS: dict[tuple[str, Any], str] = {
+    ("font", "Georgia, 'Iowan Old Style', 'Charter', 'Bitstream Charter', serif"): "Serif",
+    ("font", "system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', sans-serif"): "Sans",
+    ("font", "ui-monospace, 'Cascadia Mono', 'Source Code Pro', Menlo, monospace"): "Mono",
+    ("bg", "#ffffff"): "Light bg",
+    ("bg", "#f5e6d3"): "Sepia bg",
+    ("bg", "#1a1a1a"): "Dark bg",
+    ("fg", "#1a1a1a"): "Dark text",
+    ("fg", "#3d2914"): "Sepia text",
+    ("fg", "#e8e8e8"): "Light text",
+    ("bionic_enabled", True): "On",
+    ("bionic_enabled", False): "Off",
+    ("max_width", "55ch"): "Narrow",
+    ("max_width", "65ch"): "Standard",
+    ("max_width", "75ch"): "Wide",
+    ("max_width", "85ch"): "Extra wide",
+}
+
+
+def coerce_value(key: str, raw: str) -> Any:
+    """Coerce a Form-submitted string into the right type for the given key.
+
+    Form data arrives as strings. Most preference values are stored as
+    strings already (e.g., '18px', '#ffffff', '1.6'). The exception is
+    `bionic_enabled`, which is a boolean. Returns None if coercion fails.
+    """
+    if key == "bionic_enabled":
+        if raw == "true":
+            return True
+        if raw == "false":
+            return False
+        return None
+    return raw
+
+
+def label_for(key: str, value: Any) -> str:
+    """Return a friendly UI label for a (key, value) pair.
+
+    Falls back to str(value) when no friendly label exists. Used by the
+    sidebar template to render button text.
+    """
+    return PREFERENCE_LABELS.get((key, value), str(value))
--- a/app/services/reading/preferences.py
+++ b/app/services/reading/preferences.py
@@ -1,0 +1,66 @@
+"""Preference write helpers.
+
+`upsert_preference` is the single write path into the `preference` table.
+It enforces:
+  - read-modify-write (so the user's other preferences are preserved)
+  - "no write if the value is unchanged" short-circuit (avoids spurious
+    UPDATEs when the user clicks the same button twice)
+  - ATTRIBUTE.flag_modified so SQLAlchemy detects the in-place dict edit
+
+Per-key + per-value validation against PREFERENCE_OPTIONS happens at the
+route layer; this helper trusts its inputs.
+"""
+
+from datetime import UTC, datetime
+from typing import Any
+from uuid import UUID
+
+from sqlalchemy.orm import attributes
+from sqlmodel import Session
+
+from app.models.preference import Preference
+
+
+def upsert_preference(
+    *,
+    user_id: UUID,
+    key: str,
+    value: Any,
+    session: Session,
+) -> bool:
+    """Set `key`=`value` in the user's preferences. Returns True if a
+    write actually happened, False if the value was already what was
+    requested (the short-circuit case).
+
+    The caller is responsible for committing the session.
+    """
+    existing = session.get(Preference, user_id)
+
+    if existing is None:
+        session.add(
+            Preference(
+                user_id=user_id,
+                values={key: value},
+                updated_at=datetime.now(UTC),
+            )
+        )
+        return True
+
+    if existing.values.get(key) == value:
+        # No-op short-circuit. User clicked the same button twice; don't
+        # waste a write. The route still re-renders the fragment so the
+        # client sees the right state.
+        return False
+
+    # Merge into the existing dict. Build a NEW dict (not mutating in
+    # place) — clearer for SQLAlchemy change tracking and for any
+    # future caller that captured a reference to the old dict.
+    new_values = {**existing.values, key: value}
+    existing.values = new_values
+    existing.updated_at = datetime.now(UTC)
+    # Belt and braces: SQLAlchemy's JSON-column change detection is
+    # finicky on in-place dict edits. Even though we did `existing.values
+    # = new_values` (which is a re-assignment, not a mutation), some
+    # ORM configurations require explicit notification.
+    attributes.flag_modified(existing, "values")
+    return True
--- a/static/style.css
+++ b/static/style.css
@@ -22,6 +22,47 @@
   overflow-wrap: break-word;
 }
 
+/* Preference-toggle sidebar — opens as a Pico <details> at the bottom
+   of the reading view. Each <fieldset> is one preference key; each
+   <button> is one allow-listed option. aria-pressed reflects the
+   current value at page load. */
+.reader-controls {
+  max-width: var(--reader-max-width, 65ch);
+  margin: 2rem auto;
+}
+
+.reader-control-group {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0.5rem;
+  align-items: center;
+  margin-bottom: 1rem;
+  border: none;
+  padding: 0;
+}
+
+.reader-control-group legend {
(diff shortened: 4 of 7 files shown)


cubrox:

**Assessment:** NO-GO — Changes required before merge

This PR delivers READ-2's design well — allow-list-driven validation with a per-key + per-value gate, read-merge-write upsert, short-circuit on no-op, fragment-only response, accessible sidebar markup. CI is green, coverage is 98.65%, and the load-bearing security test (`test_css_injection_attempt_returns_422`) does what it says. However, the bionic_enabled toggle in the rendered sidebar is broken end-to-end: the button posts a value the route rejects, so clicking it returns 422. That's a user-facing functional bug on a control the ticket explicitly enumerates in DoD's "Reviewer can: open the reading view, click a button…". Holding for fix.

### Required changes (blocking)

#### 1. `bionic_enabled` button in the rendered sidebar always returns 422 on click

**File:** `templates/pages/reading.html` (and/or `app/services/reading/options.py`)

**Reproduce:**
```python
# Render the page → the bionic button HTML contains:
hx-vals='{"value": "True"}'   # (also the False button: "False")

# But coerce_value only accepts lowercase:
coerce_value("bionic_enabled", "True")   # → None
coerce_value("bionic_enabled", "False")  # → None

# So the actual POST the user makes:
POST /preferences/bionic_enabled  value=True   →  422 Invalid value
POST /preferences/bionic_enabled  value=False  →  422 Invalid value
```

**Why it slipped past the test suite:** `test_bionic_enabled_accepts_true_and_false_strings` posts the lowercase strings `"true"`/`"false"` directly to the route, bypassing the rendered template. There is no integration test that actually clicks (or simulates posting) what the rendered template emits. The "Reviewer can…" check at the bottom of the DoD would have caught this on manual verification, but the path from button → route is not pinned by any automated test.

**Root cause:** Jinja's `{{ opt | string }}` filter calls Python's `str()` on the booleans — which returns capitalised `"True"` / `"False"`. The route's `coerce_value` only accepts lowercase. Either side can move; pick one (and pin both with a test).

**Suggested fix (template side, lower risk):**
```jinja
hx-vals='{"value": "{{ (opt | string) | lower if opt is boolean else (opt | string) }}"}'
```
or even simpler — special-case the boolean key in the template:
```jinja
{% if opt is boolean %}
  hx-vals='{"value": "{{ "true" if opt else "false" }}"}'
{% else %}
  hx-vals='{"value": "{{ opt | string }}"}'
{% endif %}
```

**Suggested fix (route side, more permissive):**
```python
# in coerce_value
if key == "bionic_enabled":
    if raw.lower() == "true":
        return True
    if raw.lower() == "false":
        return False
    return None
```
The route-side fix has the side benefit of accepting either casing from any client (forms, htmx, future API consumers), at a small cost: it widens the parser slightly. Either choice is defensible; the current state where neither side handles it is the bug.

**Required test addition (regardless of which side you fix):** an integration test that fetches the rendered page, extracts the bionic button's `hx-vals` value, and posts that exact value to the route, asserting 200. Something like:

```python
def test_rendered_bionic_button_value_is_accepted_by_route(
    client: TestClient, session: Session
) -> None:
    user = _signed_in(client, session)
    passage = _make_passage(session, user.id)
    page = client.get(f"/read/{passage.id}").text
    # Extract the bionic ON button's value
    import re
    match = re.search(
        r'hx-post="/preferences/bionic_enabled"[^>]*hx-vals=\'{"value": "(True|False|true|false)"}\'',
        page,
    )
    assert match, "bionic button must render in the page"
    rendered_value = match.group(1)
    # Post EXACTLY what the rendered template emits
    response = client.post(
        "/preferences/bionic_enabled", data={"value": rendered_value}
    )
    assert response.status_code == 200
```

This is the missing pin: contract between template emission and route acceptance. Adding it now means a future template/route refactor that drifts on the casing fails the test instead of breaking the feature silently.

### What I Verified

#### Technical Requirements
- [x] All CI checks pass (lint, typecheck, test, build, actionlint, lint-agent-policies, python, Cloud Run preview deploy, etc.)
- [x] Test coverage 98.65% overall, 100% on the new modules
- [x] No new runtime dependencies (uv.lock not touched)
- [x] Code follows CLAUDE.md conventions (services under `app/services/`, routes under `app/api/`, fragments under `templates/fragments/`, page under `templates/pages/`)

#### Feature Implementation (issue #16 DoD walkthrough)
- [x] `POST /preferences/size {value: "20px"}` returns 200 with `--reader-size: 20px` and `<style id="reading-surface-style"` — `test_post_preference_returns_style_fragment_with_new_value`
- [x] Response is a fragment, not a full page — `test_response_is_a_fragment_not_a_full_page` (asserts `<html` and `<body` absent)
- [x] Preference row upserted with new value — pinned in the same test
- [x] Second toggle merges into existing values — `test_second_toggle_merges_into_existing_values`
- [x] Unknown key → 422 — `test_unknown_preference_key_returns_422`
- [x] Out-of-allow-list value → 422 — `test_value_outside_allow_list_returns_422` (posts `999px`)
- [x] Adversarial CSS injection → 422 — `test_css_injection_attempt_returns_422` (posts `red;}body{display:none;}` to /preferences/bg)
- [x] Same value twice → no DB write — `test_unchanged_value_does_not_update_updated_at`
- [x] aria-pressed reflects current values at page load — `test_sidebar_renders_with_aria_pressed_for_active_default`
- [x] Lint + type-check + tests all green
- [ ] **"Reviewer can: open the reading view, click a font-size button, see the text resize without page reload"** — works for size/font/line_height/bg/fg/max_width. **Broken for bionic_enabled** (Required change #1).

#### Code Quality Assessment

**`app/services/reading/options.py`:**
- All seven keys present (`font`, `size`, `line_height`, `bg`, `fg`, `max_width`, `bionic_enabled`).
- Per-key allow-lists are small, explicit, and CSS-ready.
- The first entry of each list matches `DEFAULT_PREFERENCES` (so the default path renders aria-pressed correctly).
- `coerce_value` handles the only non-string key (`bionic_enabled`) explicitly and returns None on any other input — exactly the right shape for the route to test against.
- `label_for` is a tidy lookup with a sensible fallback.

**`app/services/reading/preferences.py`:**
- `upsert_preference` cleanly separates the no-row-yet path (pure INSERT) from the existing-row path (read → merge → assign new dict → flag_modified). Building a NEW dict instead of mutating in place is the right call for SQLAlchemy change tracking.
- `attributes.flag_modified(existing, "values")` belt-and-braces is appropriate given JSON column quirks across SQLAlchemy versions/dialects.
- Returns a `bool` indicating whether a write happened — useful for callers and future telemetry, even though the route doesn't currently consume it.
- Docstring correctly notes the caller is responsible for committing.

**`app/api/reading.py` (new POST route):**
- Validation gates user input BEFORE any DB write or template render: key check first, coerce, then membership check. An attacker whose value fails the gate hits 422 with no DB side effects. Confirmed by inspection.
- Re-reads the row after upsert so the rendered fragment uniformly reflects the merged state — handles both new-row and existing-row cases without branching.
- Single `commit()` per call.
- Form-driven (`Annotated[str, Form()]`), HTMLResponse — matches HTMX's contract.

**`templates/fragments/reader_style.html`:**
- Single source of truth for the `<style id="reading-surface-style">` block. Page first-paint includes it; route returns it. Refactoring one updates both.
- All six CSS variables present; `| safe` is now justified by the upstream allow-list.

**`templates/pages/reading.html`:**
- Sidebar uses `<details>` for progressive disclosure (good — it's a secondary UI surface).
- `<fieldset>` + `<legend>` group preferences semantically. Screen readers announce both.
- All four required HTMX attributes present on each button (`hx-post`, `hx-vals`, `hx-target`, `hx-swap`).
- `aria-pressed="true|false"` reflects page-load state. (See suggestion below about live update.)

**`static/style.css`:**
- `.reader-controls` layout is sensible (matches reader column width).
- `aria-pressed="true"` styling via outline is distinct from Pico's hover/focus — won't get confused.

#### Architectural Compliance
- ADR-005 honored: server-rendered fragment, HTMX swap, no SPA, no new client JS.
- Reading-surface CSS-variable contract preserved: same six variable names, same `<style>` id, same `<article>` id. READ-1's contract is intact.
- READ-1's flagged dependent invariant is now the load-bearing design choice — this PR is exactly what that flag asked for.
- Stack compliance: no new deps, no new build step, no Node, no SPA imports, HTMX still CDN-loaded.
- Code organisation: `app/services/reading/` continues as the home for reading-surface logic; one module per concern.

#### Security Assessment
- **Allow-list is the gate.** The route checks BOTH `key in PREFERENCE_OPTIONS` AND `coerced_value in PREFERENCE_OPTIONS[key]`. Either alone would be insufficient; both together are correct.
- **CSS injection blocked at 422.** `test_css_injection_attempt_returns_422` posts `red;}body{display:none;}` to `/preferences/bg` and asserts 422. The malicious string never reaches the rendered `<style>` block.
- **Validation precedes I/O.** A 422 response is produced before any DB read, write, or template render. An attacker can't cause DB error states with bad input.
- **Cross-user writes:** the route's `user_id` comes from the signed-cookie session, never from the request body. No way to write another user's row.
- **Unauthenticated POST → 303 to /login**, same as the rest of the app. Pinned by `test_unauthenticated_post_redirects_to_login`.
- **Trailing-whitespace edge case is correctly rejected.** A user posting `value="20px "` with a trailing space would NOT match `"20px"` in the allow-list because `coerce_value` doesn't strip and dict membership is exact-match. That's the right strictness — anyone typing whitespace into a programmatic POST is by definition not a UI user clicking a button.

#### Performance
- Per request: 2-3 indexed lookups (passage, preference) + 0 or 1 UPDATE. Comfortably under the <100 ms perceived-latency NFR.
- Short-circuit avoids 2x DB writes when a user double-clicks the same button. Net win.
- No new external HTTP calls on the render path.

#### Testing Quality

17 tests cover the contract well **except for the template→route casing contract**. The tests pin invariants rather than surface (good): the merge test uses unmistakable values (24px / 1.8 / 75ch / 28px), the unchanged-value test asserts byte-identical updated_at, the security test posts adversarial input. The test that's missing is the one that would have caught the bionic bug: a test that posts what the rendered template actually emits.

### CI/CD Status
- Continuous Integration: SUCCESS (all jobs green)
- Cloud Run preview deploy: SUCCESS (https://pr-35---agile-flow-app-...)

1. **`aria-pressed` drifts from reality after a click.** Today the swap target is `#reading-surface-style` (the `<style>` block), not the sidebar buttons. So clicking 24px size: the CSS variable updates, the text resizes, but the buttons retain their page-load aria-pressed values. Visually, the outline marker stays on the previously-active button, even though the user's actual size is now 24px. After several clicks, the indicator drifts from reality. Options for a future polish ticket: HTMX OOB swap to also re-render the relevant fieldset, or a tiny client-side script that updates aria-pressed on the clicked button's siblings. Out of scope for READ-2; worth a follow-up issue.

2. **`upsert_preference` returns a `bool` no caller uses.** The return type is precise and informative, but the route never reads it. Either drop the return type to `None` (cleaner signature, matches actual usage), or wire it up — e.g., emit a structured log `pref_short_circuit=True` for the no-op case so we can confirm the short-circuit actually fires in production. Light preference for the latter, since the short-circuit is a perceived-latency claim worth verifying with real traffic.

3. **`coerce_value` returns `Any`.** Could tighten to `str | bool | None` for slightly stronger downstream guarantees. Strictly cosmetic given mypy is already clean.

4. **Test file's `_make_passage` is unused for ~14 of the 17 tests.** Not a bug — the route doesn't depend on a passage existing. Consider moving it to the page-rendering tests only (or factoring it out to conftest.py if it'll be reused elsewhere).

5. **The cookie-reissue bug noted in the PR description** is a known, pre-existing issue and correctly out of scope for this ticket. Worth a separate ticket to fix in middleware so all routes returning their own HTMLResponse get session re-issuance for free.

### Strengths

- **Allow-list architecture is exactly what READ-1 asked for.** The `| safe` filter on CSS values is now safe-by-construction — every value in `PREFERENCE_OPTIONS` is hand-audited CSS-ready content, and the route refuses anything else. The dependent invariant is closed.
- **Read-merge-write semantics are pinned by tests.** `test_second_toggle_merges_into_existing_values` and `test_third_toggle_updates_only_its_own_key` make it impossible for a future "let me just blow away the dict" refactor to slip through silently.
- **Short-circuit is pinned correctly.** `test_unchanged_value_does_not_update_updated_at` asserts exact equality between two consecutive `updated_at` values — bytes-identical. The inverse test (`test_changed_value_does_update_updated_at`) prevents an over-eager "always short-circuit" regression.
- **Fragment template extraction.** `templates/fragments/reader_style.html` becoming the single source of truth for both first-paint and HTMX swap is a clean factoring. The page first-paint and the toggle swap can never diverge.
- **Accessibility from day one.** `<button type="button">`, `aria-pressed`, `<fieldset>`/`<legend>`, semantic `<details>`/`<summary>`. Keyboard-first works out of the box.
- **Adversarial test is real.** `test_css_injection_attempt_returns_422` posts the kind of payload an attacker would actually craft — not a sanitised "looks bad" string. This is the right shape for a load-bearing security test.

### Files Changed Analysis

1. **`app/services/reading/options.py`** (77 lines) — allow-list, label map, coerce_value. The single source of truth for what a user can submit.
2. **`app/services/reading/preferences.py`** (66 lines) — upsert helper with read-merge-write + short-circuit + flag_modified.
3. **`app/api/reading.py`** (+72 lines) — new POST route with key+value validation, upsert, fragment response.
4. **`templates/fragments/reader_style.html`** (10 lines) — extracted style block, shared between page and toggle endpoint.
5. **`templates/pages/reading.html`** (+18 lines) — `<details>` sidebar with HTMX-attributed buttons. **Has the bionic-bool casing bug.**
6. **`static/style.css`** (+41 lines) — `.reader-controls` layout, `aria-pressed` outline marker.
7. **`tests/test_preferences.py`** (310 lines, 17 tests) — comprehensive DoD coverage; missing the template→route casing pin.

**NO-GO** — One blocking change required before merge:

1. The `bionic_enabled` button in the rendered sidebar emits `value="True"`/`value="False"` (capitalized), but the route's `coerce_value` only accepts lowercase. Clicking the bionic button always returns 422. Fix either the template (lowercase the boolean string) or the route (accept either casing), and add an integration test that posts what the rendered template actually emits to pin the contract going forward.

Once that's resolved, this PR is otherwise ready: the security architecture (allow-list as a gate before any DB or render) is correct and well-pinned, the read-merge-write upsert semantics are pinned, the short-circuit is pinned, and the accessibility markup is solid. No other blocking issues.

---

cubrox:
## Agent Review — Re-review of NO-GO from https://github.com/vibeacademy/cubrox/pull/35#issuecomment-3567145501

**Assessment:** GO — Ready for human merge

**Previous blocker status:** Addressed.

I'm a fresh reviewer with no context on the original review beyond what's posted on the thread. I read the previous NO-GO, then verified the fix commit (`b3f740a`) on its own merits.

### Was the previous blocker addressed?

**Yes.** The previous reviewer found that `{{ opt | string }}` in `templates/pages/reading.html` rendered Python `True`/`False` as the capitalized strings `"True"`/`"False"`, but `coerce_value()` only accepted lowercase `"true"`/`"false"`. The actual bionic button in the rendered UI always returned 422.

The fix takes the lower-risk template-side option the previous reviewer suggested:

1. **New helper `value_for_form()` in `app/services/reading/options.py:80-93`** that returns `"true"` / `"false"` for Python booleans and `str(value)` for everything else. The helper uses `is True` / `is False` (identity check, not truthiness) — correct, because `value_for_form(1)` would otherwise wrongly return `"true"`.

2. **Wired through to the page template context in `app/api/reading.py:79`** alongside `label_for`. Same pattern as the existing helper, no surprise.

3. **`templates/pages/reading.html:21`** — `hx-vals='{"value": "{{ value_for_form(opt) }}"}'` replaces `{{ opt | string }}`. Single-line change.

The route side (`coerce_value`) was not touched — it still requires lowercase. That's fine: there's exactly one client (the rendered template) and the helper bridges them.

### Was the integration test added that pins the contract?

**Yes** — `test_bionic_buttons_render_lowercase_booleans_and_route_accepts_them` in `tests/test_preferences.py:318-348`. It does exactly the two things the previous reviewer asked for:

- **(a) Asserts the rendered HTML contains `'"value": "true"'` and `'"value": "false"'` AND that the capitalized forms `'"value": "True"'` / `'"value": "False"'` are absent.** A future template refactor that drifts back to `{{ opt | string }}` (or any equivalent) fails both halves of (a). This is a hard pin on the rendered output.
- **(b) Round-trips both rendered values back to the route, asserts 200.** A future route change that started requiring capitalized booleans fails (b).

This is exactly the missing pin the original reviewer flagged. The test is now the contract.

A second small unit test (`test_value_for_form_helper_handles_booleans_and_strings`) exercises the helper directly with `True`, `False`, `"18px"`, `"#ffffff"`. Belt-and-braces — the integration test alone would catch regressions, but a focused unit test makes the helper's contract obvious to readers.

### Were any regressions introduced?

**No.**

- `coerce_value` is untouched — string-valued keys (`size`, `font`, `line_height`, `bg`, `fg`, `max_width`) still flow through the identity branch.
- The original 17 tests are still in place (the test file goes from 310 to 358 lines — pure addition, no test modified or deleted).
- The template change is minimal: only the boolean-handling expression changes; aria-pressed, hx-target, hx-swap, hx-post, button semantics, and fieldset structure are untouched.
- The route's allow-list gate, read-merge-write semantics, short-circuit, and fragment-only response remain unchanged.
- The helper uses `is True` / `is False` rather than truthiness, so `value_for_form(0)` still returns `"0"`, `value_for_form("")` still returns `""`, etc. No accidental coercion of non-boolean falsy/truthy values.

### CI status

**All green on the latest commit (`b3f740a`).** Verified via `gh pr checks 35`:

- CI Checks / lint: SUCCESS
- CI Checks / typecheck: SUCCESS
- CI Checks / test: SUCCESS
- CI Checks / python: SUCCESS (lint + tests + coverage; 121 tests, 98.67% — up from 119/98.65% on the previous commit, consistent with the 2 new tests at 100%)
- CI Checks / build: SUCCESS
- CI Checks / actionlint: SUCCESS
- CI Checks / lint-agent-policies: SUCCESS
- CI Checks / node: SUCCESS (correctly auto-skipped, no new JS)
- Deploy PR Preview to Cloud Run: SUCCESS

### Status of the previous reviewer's non-blocking suggestions

The previous review listed four non-blocking items: aria-pressed drift after click, `upsert_preference`'s unused bool return, `coerce_value`'s `Any` return type, and the cookie-reissue middleware bug. The author was not expected to address these in this PR. Confirmed:

- They're still present (this fix is scoped tightly to the bionic bug).
- None of them have surfaced as new blockers — they were correctly classified as polish/follow-ups.
- The fix didn't accidentally weaken any other invariant the previous reviewer praised (allow-list as a gate, read-merge-write, short-circuit, fragment-only response — all intact).

### Suggestions (non-blocking, new)

None new from this re-review. The four pre-existing non-blocking suggestions remain appropriate as follow-up tickets.

**GO** — The previous blocking issue is resolved, the integration test now pins the template→route casing contract permanently, no regressions were introduced, and CI is fully green on the latest commit. The fix is minimal, well-targeted, and matches the lower-risk approach the original reviewer suggested. Safe to merge.

---

PR #42, opened 2026-05-13
feat(auth): rate-limit /login and /auth/verify with Postgres token buckets

### LINKED ISSUE #12: AUTH-4 — Rate-limit /login and /auth/verify

cubrox:
Closes #12.

## Summary

Adds Postgres-backed token-bucket rate limiting to the two unauthenticated auth endpoints. Each request consumes one token from **two** buckets — IP and a secondary key — and is rejected with HTTP **429 + Retry-After** if either bucket is empty.

| Route | Primary key | Secondary key |
|-------|-------------|----------------|
| \`POST /login\` | \`login:ip:<ip>\` | \`login:email:<normalized-email>\` |
| \`GET /auth/verify\` | \`verify:ip:<ip>\` | \`verify:token-prefix:<8-hex-of-sha256>\` |

10 tokens per hour, linear refill (1 token / 360 s). On exhaustion the service rolls back the partial decrement and returns 429 with \`Retry-After\` set to the seconds until the next token.

## Design choices

- **Postgres, not Redis** — per the ADR: \"Redis would be better; add when traffic justifies it.\" Stays in-process. New \`rate_bucket\` table (migration 006) with three columns: \`key\`, \`tokens\`, \`refilled_at\`.
- **Two buckets per request** — the IP key blocks the single-IP spam vector. The secondary key blocks the distributed-IPs-targeting-one-victim attack: email-bombing a specific address, or brute-forcing a specific token. Both must have tokens or the request is rejected.
- **Token-prefix for verify** — secondary key for \`/auth/verify\` uses the first 8 hex chars of \`SHA-256(token)\` so the \`rate_bucket\` table never stores raw token material.
- **Cloud Run topology** — \`X-Forwarded-For\` is parsed *last-entry-first* because Cloud Run strips client-supplied entries and appends the real client IP at the end.
- **PII discipline** — the rate-limit WARN log emits a 16-hex SHA-256 prefix of the email, never the raw address. The 429 response body is a generic \"try again later\" — no email echo.
- **Test clock injection** — \`rate_limit._now_utc\` is a function reference. Tests monkeypatch it to fake clock advancement (verifies linear refill in \<1 second of real time).

## Definition of Done — Issue #12

- [x] 10 successive requests from one IP succeed, 11th returns 429
- [x] Different IPs are independently bucketed
- [x] Same email from different IPs is bucketed by email key (cross-IP defense)
- [x] \`Retry-After\` between 1 and 3600
- [x] Bucket refills linearly (verified by advancing fake clock 361 s)
- [x] 429 body does not echo the rate-limited email
- [x] WARN log line hashes the email rather than logging it raw
- [x] \`ruff check\` + \`mypy\` + \`pytest\` all pass

## Test plan

- [x] 196 tests pass locally (97.38% coverage; rate-limit service at 97%)
- [x] Migration up/down/up cycle test passes against fresh SQLite
- [ ] Manual: hammer \`/login\` 11 times with the same email + IP, observe 429 with Retry-After
- [ ] Manual: same email from 11 different IPs, observe 429 on the 11th (email bucket exhausted)
- [ ] Manual: wait 6+ minutes, hit the route again, observe 202 (linear refill works)
- [ ] Cloud Logging: trigger a 429, verify the log line shows the email-hash prefix but not the raw address

### DIFF
--- a/alembic/versions/006_add_rate_bucket.py
+++ b/alembic/versions/006_add_rate_bucket.py
@@ -1,0 +1,46 @@
+"""add rate_bucket table
+
+Revision ID: 006
+Revises: 005
+Create Date: 2026-05-12
+
+Per-key token bucket for AUTH-4 rate limiting on /login and
+/auth/verify. One row per (route, dimension, value) tuple, e.g.:
+
+  - "login:ip:203.0.113.7"
+  - "login:email:reader@example.com"
+  - "verify:ip:203.0.113.7"
+  - "verify:token-prefix:abc12345"
+
+The key is opaque to the schema; the service derives it. Tokens are
+stored as float so the linear refill math works without integer
+truncation. See app/services/rate_limit.py.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "006"
+down_revision: str | None = "005"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "rate_bucket",
+        sa.Column("key", sa.String(length=255), primary_key=True),
+        sa.Column("tokens", sa.Float(), nullable=False),
+        sa.Column(
+            "refilled_at",
+            sa.DateTime(timezone=True),
+            nullable=False,
+        ),
+    )
+
+
+def downgrade() -> None:
+    op.drop_table("rate_bucket")
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -32,6 +32,10 @@ from app.services.identity.session import (
     current_user,
     sign_session,
 )
+from app.services.rate_limit import (
+    enforce_login_rate_limit,
+    enforce_verify_rate_limit,
+)
 
 router = APIRouter()
 
@@ -47,7 +51,12 @@ GENERIC_FRAGMENT = "<p>Check your inbox for a sign-in link.</p>"
 _TOKEN_TABLE = MagicLinkToken.metadata.tables["magic_link_token"]
 
 
-@router.post("/login", response_class=HTMLResponse, status_code=202)
+@router.post(
+    "/login",
+    response_class=HTMLResponse,
+    status_code=202,
+    dependencies=[Depends(enforce_login_rate_limit)],
+)
 def login(
     background_tasks: BackgroundTasks,
     session: SessionDep,
@@ -77,7 +86,7 @@ def login(
     return GENERIC_FRAGMENT
 
 
-@router.get("/auth/verify")
+@router.get("/auth/verify", dependencies=[Depends(enforce_verify_rate_limit)])
 def verify(
     token: str,
     session: SessionDep,
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -4,6 +4,7 @@ from app.models.comprehension_question_cache import ComprehensionQuestionCache
 from app.models.magic_link_token import MagicLinkToken
 from app.models.passage import Passage
 from app.models.preference import Preference
+from app.models.rate_bucket import RateBucket
 from app.models.todo import Todo
 from app.models.user import User
 
@@ -12,6 +13,7 @@ __all__ = [
     "MagicLinkToken",
     "Passage",
     "Preference",
+    "RateBucket",
     "Todo",
     "User",
 ]
--- a/app/models/rate_bucket.py
+++ b/app/models/rate_bucket.py
@@ -1,0 +1,27 @@
+"""RateBucket model.
+
+A single row per rate-limit key (e.g. `"login:ip:1.2.3.4"`,
+`"login:email:user@example.com"`). The bucket is a classic
+token-bucket: `tokens` is the current credit balance, `refilled_at`
+is the wall-clock at the last balance update.
+
+Refill is linear: 10 tokens per hour means one token every 360
+seconds. When a request lands, we compute how many tokens have
+accrued since `refilled_at`, cap at the bucket size, decrement one,
+and write back. See `app/services/rate_limit.py` for the math.
+
+Postgres-backed (per ADR — Redis when traffic justifies). No
+foreign keys: keys are opaque strings owned by the service.
+"""
+
+from datetime import datetime
+
+from sqlmodel import Field, SQLModel
+
+
+class RateBucket(SQLModel, table=True):
+    __tablename__ = "rate_bucket"
+
+    key: str = Field(primary_key=True)
+    tokens: float
+    refilled_at: datetime
--- a/app/services/rate_limit.py
+++ b/app/services/rate_limit.py
@@ -15,10 +15,12 @@ Logging discipline: we never log raw emails. Rate-limit hits log
 `route` + 16-hex-char SHA-256 prefix of the email so support can
 correlate without leaking PII into Cloud Logging.
 
-Concurrency: production Postgres uses `SELECT ... FOR UPDATE` so two
-concurrent requests can't both decrement the same bucket below zero.
-SQLite (in tests) treats `FOR UPDATE` as a no-op, which is fine — the
-test session is single-threaded.
+Concurrency: the SELECT uses `with_for_update()` so on Postgres, two
+concurrent workers can't both read the same `tokens` value and clobber
+each other's decrement. SQLite (in tests) silently ignores the lock
+hint, which is fine — the test session is single-threaded. The
+first-touch INSERT is handled separately in `_take_token` (catches
+`IntegrityError`, recurses once into the locked-update path).
 """
 
 import hashlib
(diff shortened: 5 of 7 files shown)


cubrox:
## PR Review — #42 (AUTH-4 rate limiting)

**Reviewer:** pr-reviewer agent
**Recommendation:** CONDITIONAL — one blocking concurrency issue; the rest is clean.

### Definition of Done — Issue #12

| DoD bullet | Met? | Evidence |
|---|---|---|
| 10/IP succeed, 11th 429 | Yes | `test_ten_requests_from_same_ip_then_429_on_eleventh` |
| Different IPs independently bucketed | Yes | `test_different_ips_are_independently_bucketed` |
| Same email across IPs bucketed by email | Yes | `test_same_email_across_different_ips_is_bucketed_by_email` |
| `Retry-After` 1..3600 | Yes | `test_429_includes_retry_after_within_plausible_range` |
| Linear refill (advance clock 6 min) | Yes | `test_bucket_refills_one_token_after_six_minutes` (advances 361 s) |
| 429 body doesn't echo email | Yes | `test_429_body_does_not_leak_the_email` |
| Log line hashes email | Yes | `test_rate_limit_log_hashes_the_email`; `_hash_email` uses 16-hex SHA-256 prefix |
| ruff / mypy / pytest pass | Yes | CI run 25813883857 success; 196 tests, 97.38% coverage |

All DoD items met.

### Findings

**Required change (blocking)**

1. **Lost-update race on bucket decrement.** `app/services/rate_limit.py:90` does `session.get(RateBucket, key)` with no row-level lock, mutates `bucket.tokens`, and commits via `_enforce`. The module docstring (lines 18-21) explicitly claims "production Postgres uses `SELECT ... FOR UPDATE`" — but there is no `with_for_update()` anywhere in the file (verified with grep). Under concurrent load (Cloud Run autoscale, two simultaneous /login hits for the same email), both workers can read `tokens=5`, both write `tokens=4`, and the bucket loses a decrement. At 10/hour the practical attack surface is small, but the docstring is materially misleading and the ADR-cited concurrency guarantee is not delivered. Fix: either (a) add `session.exec(select(RateBucket).where(...).with_for_update()).one_or_none()` in `_take_token`, or (b) rewrite as a single atomic `UPDATE rate_bucket SET tokens = ..., refilled_at = ... WHERE key = ? RETURNING tokens` (`INSERT ... ON CONFLICT DO UPDATE` for first-touch). Update the docstring to match whichever path you pick.

**Suggestions (non-blocking)**

2. **Apply to `/passages/pdf` next.** PR notes acknowledge this; ticket scoped only /login and /auth/verify, so not a NO-GO. File a follow-up under area/security: PDF upload is a 25 MB resource sink without a bucket.
3. `seconds_per_token` is module-level (`SECONDS_PER_TOKEN = 360.0`); good. Consider exposing `BUCKET_SIZE` via settings later if Resend cost dynamics change.
4. The `_now_utc` indirection is clean — tests monkeypatch the symbol on the module, not a global mutable. Nice touch.

### Items scrutinized and cleared

- **Form double-read** (`enforce_login_rate_limit` calls `await request.form()`, route reads `email: Form()`): safe. FastAPI/Starlette cache `request._form` on the request scope; the route's `Form()` parser reuses it. The IP-bucket and `test_429_includes_retry_after_within_plausible_range` tests both rely on the route returning 202 after the dependency reads the form, which would fail with a 422 if the form re-parse broke.
- **Rollback semantics**: `_enforce` runs as a dependency *before* the route body, so `session.rollback()` on 429 has nothing route-side to discard. The first-touch path also adds an unpersisted `RateBucket`, which the rollback correctly evicts.
- **X-Forwarded-For trust**: `_client_ip` reads the LAST entry, matching Cloud Run topology. Cloud Run strips client-supplied XFF and appends the true client IP, so client lies are inert in production. Tests inject `"10.0.0.1, <ip>"` to mirror that.
- **Bucket-key collision** (`email=ip:1.2.3.4`): no real collision. `login:email:ip:1.2.3.4` and `login:ip:1.2.3.4` share the `login:` namespace but the second segment (`email:` vs `ip:`) disambiguates. Adding ":" inside the email value extends the key but doesn't make it equal to any IP-bucket key.
- **PII in 429 body**: `detail="Too many requests. Try again later."` is a fixed string; no email interpolation. Test 4 confirms.
- **Migration ordering**: revision 006, `down_revision = "005"`, builds on the existing chain. `test_upgrade_downgrade_upgrade_cycle` exercises 006 → 005 → 006 cleanly.
- **Token PII at /auth/verify**: secondary key uses `sha256(token)[:8]` so raw token material never lands in `rate_bucket`.

### Result

---

**Result:** CONDITIONAL — concurrency gap blocks merge until resolved
ci: pass (run 25813883857)
tests: 196 / 196, 97.38% coverage
blocking_issues: 1 (lost-update race on `_take_token`)
suggestions: 1 (follow-up for `/passages/pdf` rate limit)
dod_met: 8 / 8

cubrox:
**Blocking concurrency issue addressed in 33b0618.**

The `_take_token` function now uses `select(RateBucket).where(...).with_for_update()` so Postgres row-locks the bucket before refill + decrement. SQLite (tests) silently ignores the lock hint, which is fine — the test session is single-threaded.

The separate first-touch INSERT race (two requests both miss the SELECT, both attempt INSERT, the second hits the unique-PK constraint) is now caught with `IntegrityError → rollback → recurse once`. By the second pass the row exists and takes the locked-update path.

All 196 tests still pass (96.97% coverage; the new try/except branch is exercised in tests only when concurrent flushes collide, so coverage moved slightly).

Docstrings updated to honestly describe both code paths.

Ready for re-review when convenient.

Issue #23, opened 2026-05-02
METRIC-3 — Aggregate query for the 100k-line dashboard


cubrox:
## Standard Fields

- **Problem Statement**: Events are being persisted (METRIC-1+2) but no one can see the metric. The PRD evaluates M2 ("product-market fit") on the 100,000-line target across 3 months. We need an aggregate query exposed via a tiny CLI command that prints the running total + per-day breakdown — enough for a release decision, no UI required for MVP.
- **Parent Epic**: #7
- **Effort Estimate**: S
- **Priority**: P2

---

### A. Environment Context

- Stack: SQLModel for the query; a small CLI entrypoint via `python -m app.scripts.metric`
- Integration points: reads `reading_event` (METRIC-1)
- Existing pattern: this is the first CLI script. Set the convention: `app/scripts/<name>.py` with a `def main(): ...` and `if __name__ == "__main__": main()`. Future scripts mirror this.
- Files to create or modify:
  - CREATE `app/scripts/__init__.py`
  - CREATE `app/scripts/metric.py` (CLI entrypoint)
  - CREATE `app/services/metric/aggregate.py` (`total_lines_since(date) -> int` and `lines_per_day_since(date) -> list[(date, int)]`)
  - CREATE `tests/test_metric_aggregate.py`

### B. Guardrails

- Do NOT add an HTTP endpoint for the metric. Dashboard UI is Phase 2 per the epic. CLI-only is fine for the launch decision.
- Do NOT iterate the rows in Python. Use SQL `SUM(lines_processed)` and `GROUP BY date_trunc('day', occurred_at)` — push the aggregation to Postgres.
- Do NOT show per-user data in the CLI output. The metric is product-level (sum across all users). Per-user analytics are Phase 2.
- Do NOT pull all events into memory. The query returns aggregates only.
- The CLI takes one optional `--since YYYY-MM-DD` arg (default: 90 days ago). Output format is plain text: total + a date-sorted table.
- Do NOT swallow query errors. Let them bubble — running this against a missing table or wrong DB should fail loud, not return zero.

### C. Happy Path

1. Developer runs `uv run python -m app.scripts.metric --since 2026-08-01`.
2. Script connects to the database via the standard config.
3. Script calls `aggregate.total_lines_since(date(2026, 8, 1))` → returns single `int`.
4. Script calls `aggregate.lines_per_day_since(date(2026, 8, 1))` → returns `[(date, int), ...]` ascending.
5. Script prints:
   ```
[...]

Issue #22, opened 2026-05-02
METRIC-2 — Beacon endpoint on passage close


cubrox:
## Standard Fields

- **Problem Statement**: METRIC-1 created the table; this ticket writes to it. The reading view must emit a single `reading_event` row when the user closes the passage (navigates away or closes the tab). HTMX's `hx-trigger="unload"` plus a `sendBeacon`-style POST is the right primitive — it fires reliably on close without blocking unload.
- **Parent Epic**: #7
- **Effort Estimate**: S
- **Priority**: P1

---

### A. Environment Context

- Stack: FastAPI + HTMX 2.x; reuses `ReadingEvent` model from METRIC-1
- Integration points: triggered from the reading view (READ-1); writes to `reading_event`
- Existing pattern: HTMX-targeted POST routes from READ-2; ownership pattern from READ-1
- Files to create or modify:
  - MODIFY `app/api/reading.py` (add `POST /passages/{passage_id}/close` route)
  - MODIFY `templates/pages/reading.html` (add a hidden `<div hx-post="/passages/{{ passage.id }}/close" hx-trigger="unload from:body" hx-vals='js:{lines: window.cubroxLineCount}'>` element AND a tiny inline `<script>` that computes `cubroxLineCount` from the rendered passage `<article>` element after layout)
  - CREATE `tests/test_passage_close.py`

### B. Guardrails

- Do NOT block the unload event. Response body is irrelevant — return 204 No Content.
- Do NOT trust client-supplied `lines` blindly. Validate `1 ≤ lines ≤ 100_000`. Reject (silently — log + 204) if out of range.
- Do NOT count the same close twice. Idempotency: this ticket does NOT need a deduplication key — accept that some double-fires (refresh, browser quirks) will happen and the metric will be a slight over-count. Document this in the route docstring; revisit only if the metric materially diverges from manual counts in dogfooding.
- Do NOT log the passage text (it's not in the request body anyway, but be explicit). Log the user_id + passage_id + lines.
- Do NOT serve another user's passage close. Auth-check + ownership-check + 404 (not 403) per READ-1's convention.
- Reading-view template must NOT use any JavaScript framework (no React, Vue, etc.). The line-count snippet is plain ES2015.

### C. Happy Path

1. Authenticated user opens reading view; the inline `<script>` measures the passage `<article>` and stores `window.cubroxLineCount`.
2. User navigates away / closes the tab.
[...]

PR #46, opened 2026-05-14
feat(metric): reading_event table + Alembic migration

### LINKED ISSUE #21: METRIC-1 — reading_event table + Alembic migration
## Standard Fields

- **Problem Statement**: The PRD's primary success metric ("100,000 lines of text read/processed within 3 months") is unmeasurable without per-session event records. This ticket creates the persistence layer; METRIC-2 writes events; METRIC-3 aggregates them. Splitting these lets the migration land independently and ahead of route work.
- **Parent Epic**: #7
- **Effort Estimate**: XS
- **Priority**: P1

---

### A. Environment Context

- Stack: SQLModel + Alembic; pure DB layer
- Integration points: depends on `user` (AUTH-1) and `passage` (INGEST-1) tables for FK references
- Existing pattern: follow `app/models/passage.py` for SQLModel shape and Alembic conventions
- Files to create or modify:
  - CREATE `app/models/reading_event.py` (matches schema in [docs/TECHNICAL-ARCHITECTURE.md > Database Schema](../../docs/TECHNICAL-ARCHITECTURE.md#database-schema-sketch))
  - CREATE `alembic/versions/<timestamp>_add_reading_event.py`
  - CREATE `tests/test_reading_event_model.py`

### B. Guardrails

- Append-only by design. Do NOT add an `update_at` column. Do NOT add an UPDATE path anywhere.
- Foreign keys: `user_id REFERENCES "user"(id) ON DELETE CASCADE`, `passage_id REFERENCES passage(id) ON DELETE CASCADE`. When a user or passage is deleted, their reading events go with them — they're meaningless without context.
- Index on `occurred_at` (the dashboard query in METRIC-3 aggregates by date).
- Do NOT add an index on `(user_id, occurred_at)` yet — only add when METRIC-3 needs it. Premature indexing is waste.
- `lines_processed` is a small `INT` (≥1, ≤ ~100,000 — bounded by the paste-size cap from INGEST-1's guardrails). Use `INT NOT NULL` (no nullable, no DEFAULT — the writer in METRIC-2 always knows the count).
- Migration must be reversible. `downgrade()` drops the table cleanly.

### C. Happy Path

1. Developer runs `uv run alembic upgrade head` against a Neon branch.
2. Migration creates `reading_event` table with PK `id`, FKs to `user` and `passage`, `lines_processed INT NOT NULL`, `occurred_at TIMESTAMPTZ NOT NULL DEFAULT now()`.
3. Migration creates index on `occurred_at`.
4. Test inserts a `ReadingEvent` row, asserts it persists with the right values and a server-side default `occurred_at`.

### D. Definition of Done

- `tests/test_reading_event_model.py` asserts inserting a `ReadingEvent` row and selecting it back returns the same `user_id`, `passage_id`, `lines_processed`.
- Test asserts `occurred_at` is set to a server-side `now()` (within 1 second of test wall-clock if `occurred_at` not explicitly set).
- Test asserts `lines_processed = 0` is rejected (CHECK constraint or model validation).
- Test asserts deleting the parent `User` deletes the related `ReadingEvent` rows (CASCADE).
- Test asserts deleting the parent `Passage` deletes the related `ReadingEvent` rows (CASCADE).
- `uv run alembic upgrade head && uv run alembic downgrade -1 && uv run alembic upgrade head` runs cleanly.
- `uv run pytest` passes.
- Reviewer can: run `\\d reading_event` in psql and confirm the columns, FKs, and index match the spec.

cubrox:
Closes #21.

## Summary

Lays the persistence layer for the PRD's "100,000 lines of text processed within 3 months" success metric. This PR is XS-sized and ships nothing user-visible — but it's the foundation that METRIC-2 (#22, writes events) and METRIC-3 (#23, aggregates them) build on.

## Schema

| Column            | Type         | Constraints                                |
| ----------------- | ------------ | ------------------------------------------ |
| \`id\`              | UUID         | PRIMARY KEY                                |
| \`user_id\`         | UUID         | FK → \`user(id)\` ON DELETE CASCADE          |
| \`passage_id\`      | UUID         | FK → \`passage(id)\` ON DELETE CASCADE       |
| \`lines_processed\` | INT          | NOT NULL, CHECK >= 1                       |
| \`occurred_at\`     | TIMESTAMPTZ  | NOT NULL, DEFAULT now()                    |

Plus an index on \`occurred_at\` for METRIC-3's date-range aggregate.

## Design choices flagged

- **Append-only by design.** No \`updated_at\` column. Reading is observed, not edited.
- **CASCADE on both FKs.** A reading event without its user OR its passage is unattributable — it goes away with them. Verified against the Alembic-applied schema (the default test engine doesn't enable SQLite FK enforcement; the cascade tests turn it on explicitly).
- **CHECK >= 1 in two places.** The migration declares it, AND the SQLModel \`__table_args__\` declares it. The duplication means \`SQLModel.metadata.create_all\` (the conftest path) enforces the same invariant Postgres will — so the rejection tests work in pytest.
- **No \`(user_id, occurred_at)\` index yet.** Per ticket guardrail: premature indexing. METRIC-3 may add it once a per-user query is observed.

## Test plan

- [x] 11 new tests across two files (7 model + 4 migration)
- [x] CASCADE delete tests for both \`user\` and \`passage\` parents
- [x] CHECK rejects 0 and negative values
- [x] \`occurred_at\` defaults to now() within 1 second
- [x] Migration up/down/up cycle clean
- [x] 216 total tests pass, 96.68% coverage, ruff/mypy clean
- [ ] Reviewer can: run \`uv run alembic upgrade head\` against a Neon branch, \`\\\\d reading_event\` in psql, confirm schema

## Test coverage notes

- \`app/models/reading_event.py\` at 100%
- The CASCADE test enables SQLite's \`PRAGMA foreign_keys=ON\` via an engine event listener (per-connection, fires on every new connection from the pool). Postgres enforces FKs unconditionally; the explicit pragma is SQLite-only.

### DIFF
--- a/alembic/versions/007_add_reading_event.py
+++ b/alembic/versions/007_add_reading_event.py
@@ -1,0 +1,80 @@
+"""add reading_event table
+
+Revision ID: 007
+Revises: 006
+Create Date: 2026-05-14
+
+Append-only event table for the "100k lines processed" PRD success
+metric (Epic #7). One row per beacon dispatched on passage-close
+(METRIC-2). The aggregate query (METRIC-3) groups by `occurred_at`
+date — hence the single-column index on that column.
+
+CASCADE on both FKs: a reading event without its user or its passage
+is unattributable, so it goes away with them. Append-only — no
+`updated_at` column, no UPDATE path expected.
+
+UUID handling mirrors revisions 003 and 004: native on Postgres,
+CHAR(36) on SQLite for the in-memory test database.
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+revision: str = "007"
+down_revision: str | None = "006"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    is_pg = bind.dialect.name == "postgresql"
+    uuid_type: sa.types.TypeEngine = sa.Uuid() if is_pg else sa.String(length=36)
+
+    op.create_table(
+        "reading_event",
+        sa.Column(
+            "id",
+            uuid_type,
+            primary_key=True,
+            server_default=sa.text("gen_random_uuid()") if is_pg else None,
+        ),
+        sa.Column(
+            "user_id",
+            uuid_type,
+            sa.ForeignKey("user.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "passage_id",
+            uuid_type,
+            sa.ForeignKey("passage.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column("lines_processed", sa.Integer(), nullable=False),
+        sa.Column(
+            "occurred_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.func.now(),
+            nullable=False,
+        ),
+        # Append-only domain: a 0-line event is meaningless noise.
+        # CHECK runs in both Postgres and SQLite.
+        sa.CheckConstraint(
+            "lines_processed >= 1",
+            name="reading_event_lines_processed_positive",
+        ),
+    )
+
+    # METRIC-3's aggregate query groups by date; this is the index that
+    # makes that fast. Per the ticket: do NOT add (user_id, occurred_at)
+    # until METRIC-3 surfaces a per-user query.
+    op.create_index("ix_reading_event_occurred_at", "reading_event", ["occurred_at"])
+
+
+def downgrade() -> None:
+    op.drop_index("ix_reading_event_occurred_at", table_name="reading_event")
+    op.drop_table("reading_event")
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -5,6 +5,7 @@ from app.models.magic_link_token import MagicLinkToken
 from app.models.passage import Passage
 from app.models.preference import Preference
 from app.models.rate_bucket import RateBucket
+from app.models.reading_event import ReadingEvent
 from app.models.todo import Todo
 from app.models.user import User
 
@@ -14,6 +15,7 @@ __all__ = [
     "Passage",
     "Preference",
     "RateBucket",
+    "ReadingEvent",
     "Todo",
     "User",
 ]
--- a/app/models/reading_event.py
+++ b/app/models/reading_event.py
@@ -1,0 +1,54 @@
+"""ReadingEvent model.
+
+Append-only record of "a user finished reading some lines of a
+passage". METRIC-2 (#22) writes one row per beacon dispatched on
+passage-close. METRIC-3 (#23) aggregates them for the "100,000 lines
+processed in three months" PRD success metric.
+
+Design choices:
+  - Append-only: no `updated_at` column, no UPDATE path anywhere.
+    Reading is observed, not edited.
+  - `lines_processed` is `INT NOT NULL, CHECK >= 1` — every event
+    must claim at least one line. Zero would just be noise.
+  - `occurred_at` is server-defaulted to `now()` so the writer in
+    METRIC-2 can rely on the database's clock (consistent across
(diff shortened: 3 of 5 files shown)


cubrox:
## Code Review — PR #46 (METRIC-1, #21)

**Recommendation: GO**

Pure-DB ticket. Schema matches the PRD/architecture sketch, migration is reversible, tests exercise the full DoD, and all required CI checks (a11y, python, lint, test) are green.

### Definition of Done — verified

| DoD line | Where verified |
|----------|----------------|
| Insert/select roundtrip on (user_id, passage_id, lines_processed) | `tests/test_reading_event_model.py::test_insert_and_select_roundtrip` |
| `occurred_at` server-defaulted to `now()` within 1s | `test_occurred_at_defaults_to_now_when_unset` |
| `lines_processed = 0` rejected | `test_lines_processed_zero_is_rejected` (plus belt-and-braces `-5`) |
| CASCADE on user delete | `tests/test_reading_event_migration.py::test_delete_user_cascades_to_reading_events` |
| CASCADE on passage delete | `test_delete_passage_cascades_to_reading_events` |
| `upgrade -> downgrade -> upgrade` clean | `test_upgrade_downgrade_upgrade_cycle` |
| Required `ix_reading_event_occurred_at` index present | `test_upgrade_creates_reading_event_table` |
| Prohibited `(user_id, occurred_at)` composite index absent | Same test asserts no composite matches `{user_id, occurred_at}` |
| ruff / mypy / pytest clean | Required CI checks green |

### Schema/migration spot-checks

- **Revision chain.** 007 → down_revision 006. `ls alembic/versions/` confirms 006 (`add_rate_bucket`) was previous head; 007 is this PR's only new file. Clean.
- **UUID dialect handling.** `is_pg = bind.dialect.name == "postgresql"` branch matches the established pattern from revisions 003 and 004 — native UUID on Postgres, CHAR(36) on SQLite. Consistent.
- **CHECK duplication justified.** Constraint declared in BOTH the migration AND `SQLModel.__table_args__` under the same name (`reading_event_lines_processed_positive`). The duplication is necessary because the test conftest path uses `SQLModel.metadata.create_all` (not Alembic), so without the table-args copy the model-level rejection tests wouldn't fire. Same constraint name keeps Postgres from seeing it twice.
- **`occurred_at` defaults at both layers.** `default_factory=lambda: datetime.now(UTC)` on the model + `server_default=sa.func.now()` in the migration. Same pattern already used by `passage.created_at` and `preference.updated_at`. Consistent.
- **CASCADE test approach.** Per-engine SQLAlchemy `connect` event listener turns on `PRAGMA foreign_keys=ON` for SQLite. Postgres enforces FKs unconditionally, so this is the right way to mirror production behavior under SQLite. Cascade tests use raw SQL because (a) the SQLModel API doesn't help here — we need the DB engine to enforce, not the ORM, and (b) the event-driven pragma is per-connection on the test engine.
- **Append-only invariant.** No `updated_at` column, no UPDATE-path infrastructure. Matches the PRD design choice.

- **`lines_processed` upper bound not enforced at the schema.** The ticket text says "≤ ~100,000". The CHECK only enforces `>= 1`. I'm leaving this as a suggestion because the ticket itself describes the upper bound as "bounded by the paste-size cap from INGEST-1's guardrails" — i.e. a constraint on the *writer*, not the schema. The DB-level guarantee METRIC-2 cares about is "no zero/negative noise", which is what we have. If you want a defensive `<= 1_000_000` later, it'd go in a separate revision.

### Test count

DoD calls out 5 specific assertions; PR adds 11 across two files. Above the bar.

### Red flags scanned for

- None. No dead code, no commented-out blocks, no `# TODO` markers, no security/secrets exposure, no breaking schema change to existing tables.

---

**Result:** GO
PR: #46
Recommendation: GO
Required changes: 0
Suggestions: 1 (upper-bound CHECK on `lines_processed`, deferred to METRIC-2 author's discretion)
DoD met: 9/9
CI: a11y / python / lint / test all green (run 25873335647)

PR #40, opened 2026-05-12
fix(auth): redirect unauthenticated visitors to landing page

cubrox:
## Summary

The AUTH-3 exception handler was redirecting unauthenticated visitors to `/login`, but `/login` is a POST-only API endpoint that handles magic-link form submission — not a page. So an unauth GET to a protected route landed on a 405 Method Not Allowed instead of a sign-in form.

With the landing page at `/` (PR #39, just merged), we have a real page that serves the sign-in form. This PR points the redirect there instead. Caught by the PR #39 reviewer subagent.

## What changed

- `app/main.py`: `_handle_unauthenticated` now redirects to `/` for both browser navigation (303) and HTMX requests (HX-Redirect). Docstring updated with the why.
- 11 test assertions across 5 test files flipped from `/login` to `/`. Test names like `redirects_to_login` renamed to `redirects_to_landing` where the old name became misleading.

## Test plan

- [x] All 178 tests pass locally (97.52% coverage)
- [x] ruff check + format clean
- [x] mypy clean
- [ ] Manual: unauth GET to `/passages/new` lands on the landing page with sign-in form (not 405)
- [ ] Manual: HTMX request without session triggers HX-Redirect to `/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #47, opened 2026-05-14
feat(metric): beacon endpoint on passage close

### LINKED ISSUE #22: METRIC-2 — Beacon endpoint on passage close

cubrox:
Closes #22.

## Summary

Adds the unload-time beacon that records one \`reading_event\` row when a user leaves the reading view. With METRIC-1's table from PR #46 and this PR, the "100,000 lines read in three months" PRD success metric is now **observable** — METRIC-3 (#23) will surface the aggregate query.

## Server side

- New \`POST /passages/{passage_id}/close\` route in [app/api/reading.py](app/api/reading.py)
- Auth + ownership check (404 same body shape as cross-user / missing — no existence leak, mirrors GET /read)
- Validates \`1 ≤ lines ≤ 100_000\`; **out-of-range silently dropped** with 204 + a WARN log, because the unload path can't surface errors anyway
- Inserts one \`reading_event\` row, returns **204 No Content**

## Client side

Two template additions in [templates/pages/reading.html](templates/pages/reading.html):

1. **Inline \`<script>\`** that measures the rendered \`<article>\`'s scrollHeight ÷ computed line-height to estimate the visible-line count. Stashes the value on \`window.cubroxLineCount\`. Plain ES2015 — no framework, no build step (ADR-005).
2. **Hidden \`<div hx-trigger="unload from:body">\`** that POSTs to the new route. Uses \`hx-vals='js:{lines: window.cubroxLineCount}'\` so the count is read **at fire time**, not at render time — a preference toggle that changes line-height between page load and close still gets reflected.

## No deduplication (by design)

The route does NOT deduplicate. Refreshes, back/forward cache, browser quirks will occasionally double-fire. Per ticket guardrail, the PRD metric tolerates a small over-count for MVP; revisit only if dogfooding shows material divergence from manual line-counts.

## Note on the unauth-redirect test

The ticket DoD says "unauthenticated request returns 200 with \`HX-Redirect: /login\`". PR #40 (2026-05-12) changed the AUTH-3 redirect target from \`/login\` to \`/\` because \`/login\` is POST-only. **Tests pin the current production target (\`/\`)**, not the stale ticket reference. Same discrepancy noted in test docstring.

## Test plan

- [x] 11 new tests covering happy path, owner check, validation bounds, auth redirect (both HTMX and browser), template wiring
- [x] 227 total tests pass, 96.74% coverage, ruff/mypy clean
- [x] All four newly-enforced required checks (a11y, python, lint, test) green expected on CI
- [ ] Manual: open a reading view, close the tab, see one new \`reading_event\` row in psql with \`lines_processed\` close to the visible line count

### DIFF
--- a/app/api/reading.py
+++ b/app/api/reading.py
@@ -5,6 +5,8 @@ POST /preferences/{key}             — set one preference, return the
                                        swappable <style> fragment
 GET  /passages/{passage_id}/questions — HTMX-lazy-loaded comprehension
                                         question fragment
+POST /passages/{passage_id}/close   — unload-time beacon recording one
+                                        reading_event row (METRIC-2)
 
 Owner check on the GET: a user can only view their own passages. Other
 users' passages return 404 (not 403) — same response shape as a
@@ -20,7 +22,7 @@ import logging
 import uuid
 from typing import TYPE_CHECKING, Annotated
 
-from fastapi import APIRouter, Depends, Form, HTTPException, Request
+from fastapi import APIRouter, Depends, Form, HTTPException, Request, Response
 from fastapi.responses import HTMLResponse
 from sqlmodel import Session, select
 
@@ -28,6 +30,7 @@ from app.config import Settings, get_settings
 from app.db import get_session
 from app.models.passage import Passage
 from app.models.preference import Preference
+from app.models.reading_event import ReadingEvent
 from app.models.user import User
 from app.services.comprehension.client import get_anthropic_client
 from app.services.comprehension.generator import (
@@ -203,3 +206,67 @@ def passage_questions(
         name="fragments/questions_panel.html",
         context={"questions": questions, "error": error},
     )
+
+
+# Inclusive bounds for client-supplied line counts. 1 is the natural
+# floor (zero lines processed is meaningless noise). 100_000 mirrors
+# INGEST-1's text-length cap — any value above that means the client
+# is lying or the column count broke.
+_MIN_LINES = 1
+_MAX_LINES = 100_000
+
+
+@router.post("/passages/{passage_id}/close")
+def passage_close(
+    passage_id: uuid.UUID,
+    user: CurrentUser,
+    session: SessionDep,
+    lines: Annotated[int, Form()],
+) -> Response:
+    """Record one `reading_event` row when the user closes a passage.
+
+    Fired by HTMX `hx-trigger="unload from:body"` on the reading view's
+    hidden beacon div. Returns 204 No Content — the response body is
+    irrelevant because the browser is already tearing down the page.
+
+    Ownership check is identical to GET /read: a user closing someone
+    else's passage gets a 404 (same body as a nonexistent UUID) so
+    we don't leak which IDs exist.
+
+    Validation is permissive: out-of-range `lines` values get a 204
+    with NO row inserted (logged but otherwise silent). The unload
+    path can't surface errors anyway, so a noisy 4xx would be wasted —
+    better to drop the bad data and move on.
+
+    Idempotency: this route does NOT deduplicate. Browser quirks
+    (back/forward cache, refresh, bfcache restore) will occasionally
+    double-fire. The PRD metric tolerates a small over-count; revisit
+    only if dogfooding shows material divergence from manual counts.
+    """
+    passage = session.exec(
+        select(Passage).where(Passage.id == passage_id, Passage.user_id == user.id)  # type: ignore[arg-type]
+    ).first()
+    if passage is None:
+        # Same 404 as cross-user/nonexistent on GET /read; preserves the
+        # "existence isn't leaked" invariant.
+        raise HTTPException(status_code=404, detail="Passage not found")
+
+    if not (_MIN_LINES <= lines <= _MAX_LINES):
+        logger.warning(
+            "passage_close.invalid_lines user_id=%s passage_id=%s lines=%s",
+            user.id,
+            passage_id,
+            lines,
+        )
+        return Response(status_code=204)
+
+    session.add(
+        ReadingEvent(
+            user_id=user.id,
+            passage_id=passage_id,
+            lines_processed=lines,
+        )
+    )
+    session.commit()
+
+    return Response(status_code=204)
--- a/templates/pages/reading.html
+++ b/templates/pages/reading.html
@@ -13,6 +13,44 @@
        hx-swap="outerHTML"
        aria-live="polite"></div>
 
+  {#
+    METRIC-2: emit one reading_event on unload. The inline <script>
+    measures the rendered <article>'s height in line-heights and stashes
+    the count on `window.cubroxLineCount`. The hidden <div> below uses
+    hx-trigger="unload from:body" so HTMX fires the POST when the user
+    leaves the page (close tab, navigate away, refresh). hx-vals 'js:'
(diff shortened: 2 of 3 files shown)


cubrox:
## PR Review — #47 (METRIC-2)

**Recommendation: GO** (humans merge — I do not approve via `gh pr review --approve`).

### Definition of Done — all checked

| DoD item | Status | Evidence |
|---|---|---|
| `lines=42` → 204 + row with right user/passage/lines | OK | `test_close_inserts_reading_event_and_returns_204` |
| Row's `user_id` / `passage_id` match URL + caller | OK | same test asserts both fields |
| Other user's passage → 404, no row | OK | `test_close_on_other_users_passage_returns_404_no_row` |
| `lines=0/-1/100001` → 204, no row | OK | three dedicated tests |
| Unauthenticated → HX-Redirect, no row | OK | `test_close_unauthenticated_redirects_no_row` (target pinned to `/` per PR #40; rationale documented in docstring) |
| Template contains beacon div + correct URL | OK | `test_reading_view_template_contains_close_beacon` |
| `ruff` / `mypy` / `pytest` clean | OK | CI green, all 9 required checks pass on run 25873796064 |

### Server-side findings

- **Owner-check consistency:** the new route uses the identical pattern as GET /read (`select(Passage).where(Passage.id == passage_id, Passage.user_id == user.id).first()`) and raises the same 404 with the same body. The "no existence leak" invariant from READ-1 is preserved.
- **Validation bounds:** both inclusive endpoints (1 and 100_000) are pinned with persistence tests; 0 / -1 / 100_001 are pinned silent-reject. Boundary coverage is tight.
- **Silent-rejection semantics:** 204 carries no body (`assert response.text == ""`), table is empty (`session.exec(select(ReadingEvent)).all() == []`), WARN log includes `user_id` + `passage_id` + `lines` — no passage text in the log line. Guardrail respected.
- **Race with unload:** Starlette buffers the response before sending, so `session.commit()` runs to completion server-side regardless of whether the client is still listening. No data-loss concern.
- **No deduplication:** explicitly documented in the route docstring AND the PR description. Acceptable for MVP; revisit if dogfooding shows material divergence.

### Client-side findings

- **Line-count measurement:** `Math.max(1, Math.round(surface.scrollHeight / lineHeight))` is reasonable. `scrollHeight` is correct for rendered content height; `Math.max(1, ...)` matches the server's `_MIN_LINES`. The fallback to `fontSize * 1.2` when `lineHeight` is "normal" is the right call. **Minor caveat:** very long unwrapped lines that wrap many visual rows will over-count line *visuals*, not source-lines — given the over-count tolerance already documented, not blocking.
- **`hx-vals 'js:'` evaluation timing:** fires at unload time, so a between-load preference toggle (which changes line-height) is correctly reflected. Intended behavior.
- **Hidden `<div>` a11y:** `aria-hidden="true"` + `style="display:none"` correctly removes it from the AT tree. The a11y CI job (now gated) passed.

1. **Future hardening: prefer `pagehide` over `unload`.** MDN flags `unload` as increasingly unreliable (Safari often won't fire on tab close, Chrome's bfcache breaks it). The ticket explicitly chose `unload`, so this stays out of scope for this PR — log as a follow-up for METRIC hardening.
2. **422 on non-integer `lines`:** FastAPI/Pydantic will coerce or 422 a string `lines`. Unload path won't send strings normally, so not worth a test, but worth noting if a future client ever calls this route directly.
3. **No-JS fallback:** beacon doesn't fire without JS — but the reading view's preference toggles need HTMX anyway, so the surface is already JS-required. Acceptable.

### Red flags scanned

- No secret/token logging — confirmed.
- No SQL string interpolation — uses SQLModel `select().where()`.
- No new dependencies.
- No auth bypass — `CurrentUser` dependency present and enforced.
- No PII or passage text in logs — confirmed.

---

**Result:** GO — DoD fully met, CI green, owner-check pattern consistent with READ-1, validation pinned at boundaries, no-dedup tolerance explicitly documented. Human approves and merges.

PR #102, opened 2026-05-25
test(a11y): restore Playwright seed via Supabase admin (#97)

cubrox:
## Summary

Restores the 4 reading-surface accessibility tests that were skipped in SUPA-2c (#91) when the legacy test-seed router was deleted. Closes #97.

The new router uses Supabase's admin API to create a confirmed user without an email round-trip, then exchanges a known password for a real session JWT — same shape `/auth/callback` produces, so `current_user` accepts it identically.

## What changes

- **`app/api/test_seed.py`** (new) — `POST /test/seed-passage-and-login`. Creates a throwaway Supabase user via `service_client().auth.admin.create_user`, signs in with the random password, then seeds a Passage (+ optional Preference) row owned by the Supabase user UUID. Two-layer guard preserved (module-level `RuntimeError` + conditional include in `app/main.py`).
- **`app/main.py`** — conditional `include_router(test_seed.router)` gated on `CUBROX_TEST_SEED_ENABLED=true`.
- **`tests/test_test_seed_guard.py`** (new, 3 tests) — pins the import-time guard: rejects non-test env, accepts `CUBROX_TEST_SEED_ENABLED=true`, accepts `ENVIRONMENT=test`.
- **`tests/test_seed_route.py`** (new, 9 tests) — variant matrix, cookie shape, admin-API call shape, password-grant call shape, 422 on unknown variant, independent users across calls, 500 when Supabase returns no session.
- **`tests/a11y/reading_surface.spec.ts`** — removed the `test.skip(true, …)` SUPA-2c stub. All 4 variants now run.
- **`tests/a11y/playwright.config.ts`** — forwards `SUPABASE_URL` / `SUPABASE_ANON_KEY` / `SUPABASE_SERVICE_KEY` to the uvicorn subprocess **only when set in `process.env`**, so local dev's `.env`-based config isn't overridden with empty strings.
- **`.github/workflows/ci.yml`** — a11y job now installs Supabase CLI, runs `supabase start` (Docker stack, ~30s), and exports the local keys via `$GITHUB_ENV` for the Playwright step to forward.

## Why the design

- **`admin.create_user` + `sign_in_with_password`** instead of `admin.generate_link` because the password-grant path returns a ready-to-use JWT directly. `generate_link` returns a URL that still requires the hash-fragment callback dance to exchange — extra moving parts for no benefit.
[...]

PR #100, opened 2026-05-25
docs+chore: drop dead deps + write ADR-006 + docs (SUPA-6, #85)

va-worker:
## Summary

Migration epic (#79) closeout. Strips dead dependencies, writes ADR-006 superseding ADR-002, adds a Pattern Library entry for the Supabase hash-fragment callback, and reconciles every reference doc with the post-migration reality.

## Deps

```diff
- "alembic>=1.13",
- "itsdangerous>=2.2.0",
- "resend>=2.29.0",
```

`psycopg` **stays** — SQLAlchemy needs it as the Postgres driver for `app/db.py`'s engine. The SUPA-6 ticket draft listed it for removal but that would crash the app on startup.

`uv tree | grep -E "resend|itsdangerous|alembic"` returns nothing after the lock regen.

## Docs

**`docs/TECHNICAL-ARCHITECTURE.md`:**
- Authentication section rewritten for the Supabase flow (`sign_in_with_otp`, two-stage callback, `sb-access-token` cookie, `auth.users` as identity source)
- Database section rewritten: Neon → Supabase Postgres, Alembic → Supabase CLI migrations, RLS as the defensive layer
- ADR-002 marked **Superseded by ADR-006**; original text preserved (ADRs are append-only history)
- **NEW: ADR-006** captures the context (Resend test-mode failure that triggered the migration), the decision, and the consequences

**`docs/PATTERN-LIBRARY.md`:**
- **NEW: Pattern #35** — Supabase Hash-Fragment Callback Requires a Two-Stage Handler. Code shape + security trade-off + most-common breakage (missing allowlist entry)

**`CLAUDE.md`:**
- Database layer + Database swapped to Supabase
- "Email: Resend" removed; "Auth: Supabase Auth" added
- Project Board URL corrected (`users/cubrox/projects/1` was wrong — cubrox is a repo not a user; now points at `orgs/vibeacademy/projects/29`)

**`app/db.py`:** docstring + inline comments stop referencing Neon (point at Supabase's Supavisor pooler).

## Test plan

- [x] `uv run ruff check .` clean
- [x] `uv run ruff format . --check` clean (67 files already formatted)
- [x] `uv run mypy app/` clean (41 source files)
- [x] `uv run pytest` — 198 passed
- [x] `uv tree | grep -E "resend|itsdangerous|alembic"` returns nothing

## What happens at merge

- Production gets a deploy with the smaller dependency set (no functional change since none of the removed packages were being imported)
[...]

PR #104, opened 2026-05-26
fix(auth): coerce Supabase user.id to uuid.UUID at boundary (#103)

### LINKED ISSUE #103: BUG-3 — Paste and PDF upload return 500 Internal Server Error in production
## Standard Fields

- **Problem Statement**: Both `POST /passages` (paste-text) and `POST /passages/pdf` (upload) currently return 500 to the user in production. The product's core ingestion flow is broken end-to-end — a signed-in user can authenticate but cannot get a passage into the reading view. Discovered by Laura testing the live site on 2026-05-25, after the Supabase migration epic (#79) shipped 2026-05-23 → 2026-05-25.
- **Effort Estimate**: S (1–4h) — likely a single root cause once Cloud Run logs are read; could escalate to M if the failure is layered.
- **Priority**: P0 — blocks the entire MVP flow. Cannot share the URL with a real user until fixed.

---

### A. Environment Context

- Stack: FastAPI + supabase-py + SQLModel on Cloud Run (per `docs/TECHNICAL-ARCHITECTURE.md`)
- Affected routes:
  - `POST /passages` — `app/api/passages.py:62`
  - `POST /passages/pdf` — `app/api/passages.py:99`
- Identity layer: `app/integrations/supabase/auth.py` — `current_user` dependency returns a Supabase user object; routes read `user.id` (UUID string)
- Data layer: `Passage.owner_id: uuid.UUID` (`app/models/passage.py`) — renamed from `user_id` in SUPA-2c (#91)
- Production logs: `gcloud logging read 'resource.type=\"cloud_run_revision\" severity>=ERROR'` (operator access required)
- Likely-suspect zones (in priority order):
  1. The SUPA-2c `user_id → owner_id` rename — any consumer it missed
  2. Supabase user object's `.id` shape (string) vs. what SQLModel/Pydantic expects for `uuid.UUID` field
  3. Session/transaction state in either handler
  4. Production-only env-var or secrets mounting (less likely — deploy.yml correctly injects `SUPABASE_*` from Secret Manager)
- Files likely to be modified: `app/api/passages.py`, possibly `app/integrations/supabase/auth.py`
- Tests likely to be added: regression test in `tests/test_passages.py` reproducing the production-shape failure

---

### B. Guardrails

- Do NOT patch the symptom (e.g. wrap the handler in try/except and return a friendly error) without finding the root cause. The user-facing error is the symptom; fix the cause.
- Do NOT change the `Passage` schema or migrations to work around the bug — schema is canonical after SUPA-2c.
- Do NOT remove or disable `current_user` to test — verify the user object's actual shape via a print/log in a preview deploy or local repro.
- Performance: no change to existing latency profile.
- The fix MUST include a pytest regression that fails before the fix and passes after — this exact failure mode escaped CI, so we need test coverage that would have caught it.

---

### C. Happy Path (post-fix)

1. Signed-in user submits `POST /passages` with form field `text=\"…some text…\"`.
2. `current_user` resolves the Supabase JWT cookie to a `gotrue.types.User` with `.id` and `.email`.
3. Handler constructs `Passage(owner_id=user.id, …)` — `user.id` coerces to `uuid.UUID` cleanly.
4. `session.add(passage); session.commit()` — INSERT into Supabase Postgres `passage` table succeeds with the user's UUID as `owner_id`.
5. Handler returns 303 redirect to `/read/{passage.id}`.
6. Same flow for `POST /passages/pdf` with the parsed text instead of the form field.

---

### D. Definition of Done

- Cloud Run logs read; root cause identified and named in the PR description (not \"fixed mysteriously\").
- Fix committed; the original repro path returns 303 (not 500) end-to-end on a preview deploy.
- New regression test in `tests/test_passages.py` that:
  - Reproduces the production-shape failure (likely involves the real Supabase user object shape, not just a SimpleNamespace stand-in)
  - Fails before the fix, passes after
- `uv run pytest`, `uv run ruff check .`, `uv run mypy app/` all clean.
- Manual verification on a preview deploy: paste a passage → land on `/read/{id}` → reading view renders. Same for PDF upload.
- Optional: if the root cause was a type/shape mismatch the test fixtures missed, file a follow-up to tighten the `conftest.py` Supabase user stub to match `gotrue.types.User` more faithfully.

cubrox:
## Summary

Fixes the production 500 on paste + PDF upload reported in #103. Root cause: Supabase returns `user.id` as a string UUID; SQLModel `table=True` bypasses Pydantic validation; SQLAlchemy then crashes on `.hex` access at INSERT. Fixed at the `current_user` boundary so every downstream consumer automatically gets the right type.

Closes #103.

## Root cause (named, not "fixed mysteriously")

```
File ".venv/lib/python3.12/site-packages/sqlalchemy/sql/sqltypes.py:3734"
    value = value.hex
            ^^^^^^^^^
AttributeError: 'str' object has no attribute 'hex'
[SQL: INSERT INTO passage (id, owner_id, text, ...) VALUES (?, ?, ?, ...)]
```

The chain:

1. `current_user` returned `resp.user` directly — a `gotrue.types.User` whose `.id` is a string (Supabase's wire format).
2. Route handlers passed this directly: `Passage(owner_id=user.id, ...)`.
3. `Passage` is a SQLModel `table=True` model. **SQLModel `table=True` bypasses Pydantic field validation at construction** — `.owner_id` stayed a string instead of being coerced to `uuid.UUID`.
4. At INSERT, SQLAlchemy's UUID column type called `value.hex` to serialize. Strings don't have `.hex` → AttributeError → 500.

Tests passed because `tests/conftest.py::make_user()` returns `SimpleNamespace(id=uuid.uuid4(), ...)` — already a `uuid.UUID` instance, so the SQLAlchemy serialization path was never exercised against the production shape.

## Why this PR's scope is bigger than the reported symptom

The reported symptom was paste + upload (POST /passages, POST /passages/pdf). But the same string-vs-UUID mismatch affects EVERY `owner_id` write site:

| File | Site |
|------|------|
| `app/api/passages.py:84` | paste INSERT (reported) |
| `app/api/passages.py:146` | PDF upload INSERT (reported) |
| `app/api/reading.py:132` | preference upsert |
| `app/api/reading.py:264` | reading event INSERT |

Plus all the SELECTs that compare `Passage.owner_id == user.id` — silently returns empty when comparing UUID column to string.

## Why this PR fixes at the boundary, not at the call sites

Three options were considered:

1. **Per-call-site coercion** — `owner_id=uuid.UUID(str(user.id))` × 4. Works but scatters the trap; the next contributor adding a 5th call site will hit it again.
2. **SQLModel field validator** — would fix at the model. But `table=True` doesn't reliably run Pydantic validators on assignment, so this is fragile.
3. **Boundary fix at `current_user`** — `AuthenticatedUser` dataclass that forces `.id: uuid.UUID`. One coercion point; type system enforces it for every consumer. **Chosen.**

## What changes

- **`app/integrations/supabase/auth.py`** — adds `AuthenticatedUser` frozen dataclass with `id: uuid.UUID` and `email: str`. `current_user` now returns `AuthenticatedUser` (not `Any`); `try_current_user` returns `AuthenticatedUser | None`. The two `Annotated[…, Depends(…)]` re-exports updated to match.
- **`tests/test_paste_production_shape.py`** (new, 4 tests) — pins the fix:
  - `test_paste_succeeds_when_supabase_user_id_is_string` — the reported failure. Fails before the fix (500), passes after (303).
  - `test_authenticated_user_dataclass_coerces_string_to_uuid` — unit-pin on the type contract.
  - `test_current_user_returns_authenticated_user_with_uuid_id` — boundary behavior via `/api/me`.
  - `test_paste_via_dependency_override_still_works_with_uuid_id` — regression guard for the existing `signed_in()` test pattern.

## Test plan

- [x] `uv run pytest` — 214 passed (4 new). Pre-fix run shows the reported failure exactly: `AttributeError: 'str' object has no attribute 'hex'`.
- [x] `uv run ruff check . && uv run ruff format --check .` — clean
- [x] `uv run mypy app/` — clean (the `Any` → `AuthenticatedUser` type change had no callers to update; routes were already duck-typed)
- [x] Pre-push hook passed
- [ ] CI green on this PR
- [ ] Manual verification on a preview deploy: paste a passage → land on `/read/{id}`. Same for PDF upload.

## Why this escaped CI before now

The CI a11y job (added in #102) DID write through this path — but its test_seed router uses `service_client().auth.admin.create_user` + signs in with a generated password, and the resulting user object is the real `gotrue.types.User`. I had pre-coerced its id with `uuid.UUID(str(auth_resp.user.id))` in `app/api/test_seed.py` — which is why the a11y CI passed at 3m17s green. So a11y CI was implicitly defending against this exact bug, but only inside the seed route. The bug was sitting in `current_user` itself, exposed on every authenticated request handler in production.

## Suggested follow-up (non-blocking)

The `conftest.py::make_user()` helper still returns a `SimpleNamespace`. Consider updating it to return an `AuthenticatedUser` for better type-realism in tests. Not in scope here because it'd require changing every `signed_in()` consumer and the tests pass cleanly with the existing pattern.

### DIFF
--- a/app/integrations/supabase/auth.py
+++ b/app/integrations/supabase/auth.py
@@ -6,14 +6,18 @@ The `current_user` / `try_current_user` FastAPI dependencies and the
 of identity — the legacy itsdangerous cookie path was deleted along
 with the User SQLModel + MagicLinkToken table.
 
-The Supabase user object (returned by `db.auth.get_user(token)`) is
-the canonical identity now. Routes that previously took
-`Annotated[User, Depends(current_user)]` now get a `gotrue.types.User`
-or equivalent — the only attributes most routes touch are `.id` and
-`.email`, which the Supabase shape preserves.
+`current_user` returns an `AuthenticatedUser` dataclass — a thin
+adapter over the raw Supabase user object that normalizes `.id` from
+the Supabase string-UUID shape to a real `uuid.UUID` instance. Without
+the normalization, SQLModel `table=True` fields with `uuid.UUID` type
+silently accept strings at construction time, then SQLAlchemy's UUID
+column type crashes on `.hex` access at INSERT — surfacing to the
+user as a 500. See #103.
 """
 
-from typing import Annotated, Any
+import uuid
+from dataclasses import dataclass
+from typing import Annotated
 
 from fastapi import Depends, Request
 
@@ -32,13 +36,28 @@ class UnauthenticatedError(Exception):
     """
 
 
-def current_user(request: Request) -> Any:
+@dataclass(frozen=True)
+class AuthenticatedUser:
+    """App-facing identity object with normalized types.
+
+    Wraps the raw `gotrue.types.User` returned by Supabase, exposing
+    only the two attributes the rest of the app reads (`.id`, `.email`)
+    and forcing `.id` to a `uuid.UUID` instance (Supabase returns a
+    string). The route handlers' SQLModel writes depend on this — see
+    module docstring for the failure mode this prevents.
+    """
+
+    id: uuid.UUID
+    email: str
+
+
+def current_user(request: Request) -> AuthenticatedUser:
     """FastAPI dependency that resolves the request's Supabase user.
 
     Reads the `sb-access-token` cookie, validates it via Supabase
-    Auth, and returns the Supabase user object. Raises
-    `UnauthenticatedError` on any failure (no cookie, invalid token,
-    Supabase unreachable).
+    Auth, and returns an `AuthenticatedUser` with `.id` coerced to
+    `uuid.UUID`. Raises `UnauthenticatedError` on any failure (no
+    cookie, invalid token, Supabase unreachable).
     """
     sb_token = request.cookies.get(SUPABASE_COOKIE_NAME)
     if not sb_token:
@@ -49,10 +68,14 @@ def current_user(request: Request) -> Any:
         raise UnauthenticatedError() from exc
     if resp is None or resp.user is None:
         raise UnauthenticatedError()
-    return resp.user
+    sb_user = resp.user
+    return AuthenticatedUser(
+        id=uuid.UUID(str(sb_user.id)),
+        email=sb_user.email or "",
+    )
 
 
-def try_current_user(request: Request) -> Any | None:
+def try_current_user(request: Request) -> AuthenticatedUser | None:
     """Soft-auth variant: returns `None` instead of raising on failure.
 
     Used by routes that render differently for signed-in vs anonymous
@@ -67,5 +90,5 @@ def try_current_user(request: Request) -> Any | None:
 
 
 # Re-export Depends-annotated aliases for convenience at call sites.
-CurrentUser = Annotated[Any, Depends(current_user)]
-OptionalUser = Annotated[Any | None, Depends(try_current_user)]
+CurrentUser = Annotated[AuthenticatedUser, Depends(current_user)]
+OptionalUser = Annotated[AuthenticatedUser | None, Depends(try_current_user)]
--- a/tests/test_paste_production_shape.py
+++ b/tests/test_paste_production_shape.py
@@ -1,0 +1,137 @@
+"""Regression test for #103: paste + upload 500 in production.
+
+The pre-fix bug:
+  - The existing `signed_in()` test helper returns `SimpleNamespace(id=uuid.uuid4(), ...)`
+    where `.id` is a `uuid.UUID` instance.
+  - Production's Supabase user object returns `.id` as a STRING UUID.
+  - SQLModel `table=True` models bypass Pydantic validation on field
+    assignment, so `Passage(owner_id=user.id)` with a string `user.id`
+    silently stores the string.
+  - SQLAlchemy's UUID column type then calls `.hex` on the value at
+    INSERT time → `AttributeError: 'str' object has no attribute 'hex'`
+    → FastAPI 500.
+
+The fix:
+  - `current_user` (in `app/integrations/supabase/auth.py`) now returns
+    an `AuthenticatedUser` dataclass that coerces `.id` to `uuid.UUID`
+    at the boundary. Every downstream call site automatically receives
+    the right type without per-route changes.
+
+This test reproduces the production shape (string `.id`) by going
+through the real `current_user` dependency with a mocked Supabase
+client that returns the string shape. Without the fix it 500s; with
+the fix it 303s.
+"""
+
+import uuid
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+from fastapi.testclient import TestClient
+from sqlmodel import Session, select
+
+from app.integrations.supabase.auth import AuthenticatedUser, current_user
+from app.main import app
+from app.models.passage import Passage
+
+
+@pytest.fixture
+def supabase_user_with_string_id(supabase_mock: MagicMock) -> str:
+    """Wire supabase_mock so `current_user` resolves to the production
+    shape (string `.id`, string `.email`) when given any cookie value.
+
+    Returns the string-uuid we configured the mock with so tests can
+    assert ownership.
+    """
+    uid_str = "00000000-0000-0000-0000-0000abcdef01"
+    supabase_mock.auth.get_user.return_value = SimpleNamespace(
+        user=SimpleNamespace(id=uid_str, email="real@example.test"),
+    )
+    return uid_str
+
+
+def test_paste_succeeds_when_supabase_user_id_is_string(
+    client: TestClient,
+    session: Session,
+    supabase_user_with_string_id: str,
+) -> None:
+    """The reported production failure: POST /passages 500s when
+    `current_user.id` is a string. With the AuthenticatedUser adapter
+    in current_user, it coerces to uuid.UUID and the INSERT succeeds.
+    """
+    # Set the Supabase cookie so current_user actually runs (don't use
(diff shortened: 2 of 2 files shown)


cubrox:
## PR Review — #104

> **Conflict of interest disclosure:** This PR was authored by the same Claude session running this review. Per the `feedback_review_consistency` lesson (and yesterday's session journal entry on #100's first review missing 8 required changes), I checked extra carefully for things the author would not have noticed — including re-running the regression test against a pre-fix version of the file to verify it actually discriminates.

### Requirements
- Fix #103 (paste + PDF upload 500) at root cause, not symptom. **Verified** — root cause (`SQLModel table=True bypasses Pydantic validation` → string `owner_id` → SQLAlchemy `.hex` crash) is named and accurate. I independently reproduced the chain in an isolated SQLModel + UUID column test and got the exact same `AttributeError`.
- Regression test fails before the fix, passes after. **Verified** — I restored `current_user` to its pre-fix body (`return resp.user`) and re-ran the suite. `test_paste_succeeds_when_supabase_user_id_is_string` failed with the production `AttributeError: 'str' object has no attribute 'hex'` at INSERT, matching the deployed bug exactly. With the fix in place: 4/4 pass.
- No regression risk on existing flows. **Verified** — grepped every `user.id` / `user.email` / `user.*metadata` access across `app/`, `tests/`, and `templates/`. Only `.id` and `.email` are read anywhere. No code paths depend on `gotrue.types.User`-only attributes (`.user_metadata`, `.app_metadata`, `.created_at`, etc.). The boundary swap is safe. Full suite: 214/214 pass.
- CI green. **Verified** — all 19 checks pass including a11y (2m26s) and the Cloud Run preview deploy.
- Closes #103. **Verified** — PR body has `Closes #103`, so the bug will auto-close on merge.

### Findings

The fix itself is correct and minimal: one coercion point at the dependency boundary, which is the right architectural choice (Option 3 in the PR body) over scattering `uuid.UUID(str(...))` at every write site. Type system now enforces the contract at every consumer.

Three non-blocking observations follow. None blocks the merge — the load-bearing test is solid and the production bug will not recur.

### Recommendation: GO

### Required Changes
None.

1. **Two of the four new tests are vacuous — they pass even without the fix.** I verified this by reverting `current_user`'s body. With the pre-fix code (returning `resp.user` directly), only `test_paste_succeeds_when_supabase_user_id_is_string` fails; the other three still pass:
   - `test_authenticated_user_dataclass_coerces_string_to_uuid` — constructs an `AuthenticatedUser` with a real `uuid.UUID` and asserts `isinstance`. The docstring claims it pins "the dataclass type annotation isn't enough to coerce on its own; current_user's job is to do the coercion," but the test never tries the string path. Frozen dataclasses do NOT enforce type annotations at runtime (verified: `AuthenticatedUser(id='abc', email='x')` stores the string without error). The test is honest in its inline comment but the name and docstring overpromise.
   - `test_current_user_returns_authenticated_user_with_uuid_id` — uses `"00000000-0000-0000-0000-0000abcdef01"` (already lowercase). Both `string == string` and `str(uuid.UUID(string)) == string` evaluate True for this input, so the assertion is satisfied whether or not coercion happened. The author's own inline comment acknowledges this. A fix would be to assert `isinstance(...)` on the returned object's `.id` (which would require exposing it differently, since `/api/me` does `str()` for JSON serialization), OR to use a mixed-case input where the comparison would discriminate.

   Filing as Suggestion rather than Required because the one working test (`test_paste_succeeds_when_supabase_user_id_is_string`) genuinely covers the production failure path end-to-end and would catch any boundary regression — the bug cannot recur as long as that test exists. The two weak tests just don't earn their docstrings.

2. **The new `AuthenticatedUser` type doesn't fully propagate.** Four route modules still have local re-declarations:
   - `app/api/auth.py:41` — `CurrentUser = Annotated[object, Depends(current_user)]`
   - `app/api/home.py:30` — `OptionalUser = Annotated[Any | None, Depends(try_current_user)]`
   - `app/api/reading.py:60` — `CurrentUser = Annotated[Any, Depends(current_user)]`
   - `app/api/passages.py:33` — `CurrentUser = Annotated[Any, Depends(current_user)]`
   
   These shadow the new strongly-typed `CurrentUser`/`OptionalUser` aliases the PR added to `auth.py:93-94`. Result: route handlers still see `user: Any` from mypy's perspective and lose the type-system enforcement the boundary fix is meant to provide. A future contributor adding a 5th route could regress to a string-shape consumer and mypy wouldn't catch it. Cheap follow-up: delete the four local aliases and import from `app.integrations.supabase.auth`. Not in scope here because the runtime fix is independent of the type-checking signal.

3. **Only paste is covered by a production-shape regression test.** Issue #103 covers both paste AND PDF upload, plus the two `reading.py` write sites (`upsert_preference` at line 132, `ReadingEvent` insert at line 264) carry the same string-vs-UUID failure mode. The boundary fix solves all four, but only `POST /passages` is exercised with the string-id production shape. Since the fix is structural (one coercion at the dependency), a single smoke test arguably suffices — but the issue's DoD asked for a test that "reproduces the production-shape failure" and the upload path is one production failure, not just paste. Cheap follow-up: parameterize the existing test or add three more thin wrappers. Filing as Suggestion because the fix being at the boundary makes the additional coverage low-marginal-value.

### Trust note

I am the same model that wrote this PR. I went looking specifically for the things the author would not have flagged — vacuous tests, propagation gaps, scope-gap between issue DoD and test coverage. I found three, and all three are non-blocking. The core fix is correct, the root cause is real, and the load-bearing test does discriminate. **GO with no required changes.**

---

**Result:** GO
required_changes: 0
suggestions: 3

Issue #81, opened 2026-05-23
SUPA-2 — DB migration: Alembic → Supabase SQL + RLS


tck517:
## Standard Fields

- **Problem Statement**: Set up the Supabase Postgres schema for the cubrox app tables (passages, reading events, etc.) with `owner_id` FKs to `auth.users` and Row-Level Security policies, so the anon key is safe to use from request handlers. This ticket is **additive-only** — it creates the new schema on Supabase without removing the legacy Neon schema or any Python code. The legacy code is dropped in SUPA-2b (#87) after SUPA-3 (#82) swaps the auth layer.
- **Parent Epic**: #79
- **Effort Estimate**: M  (reduced from L; destructive bits split to #87)
- **Priority**: P1

---

### A. Environment Context

- Stack: Supabase Postgres + Supabase CLI migrations (raw SQL)
- Integration points: Supabase remote (project ref `gnswmcgaztcxslirulwm`), Supabase GitHub auto-branching (verifies the migrations on a per-PR branch)
- Existing patterns to follow:
  - `~/projects/feature-x/leadflo/supabase/migrations/20251228160207_initial_schema.sql` for the `auth.users` FK pattern
  - `~/projects/feature-x/leadflo/supabase/migrations/20251228160242_rls_policies.sql` for RLS policy shape
  - `alembic/versions/001-007` for the schema-of-record to translate from
- Files likely to be created or modified:
  - CREATE `supabase/migrations/<timestamp>_initial_schema.sql` — all app tables with `owner_id UUID REFERENCES auth.users(id) ON DELETE CASCADE`
  - CREATE `supabase/migrations/<timestamp>_rls_policies.sql` — RLS for every tenant-scoped table
- Depends on: SUPA-1 (#80) — needs Supabase project + CLI

### B. Guardrails

- Do NOT touch any Python code, `app/models/*.py`, `app/db.py`, or `alembic/`. That's SUPA-2b (#87).
- Do NOT translate the schema verbatim — review each table. Some columns/indexes designed around our owning the `user` table may be obsolete in the `auth.users`-FK world.
- Do NOT skip RLS on any table containing user-scoped data. An unprotected table + anon key is a public table.
- Use `supabase migration new <name>` to scaffold migration files — never invent filenames or rely on memory for format.
[...]

PR #106, opened 2026-05-26
fix(db): rename user_id → owner_id on production tenant tables (#105)

### LINKED ISSUE #105: BUG-4 — Production schema still has user_id (column owner_id of relation passage does not exist)
## Problem Statement

After #104 merged, the live site moved past the SQLAlchemy `.hex` crash and hit the NEXT layer's failure: `psycopg.errors.UndefinedColumn: column \"owner_id\" of relation \"passage\" does not exist`. The Supabase migrations in `supabase/migrations/*.sql` were written with `owner_id` (post-SUPA-2c shape), but the production Supabase Postgres still has the alembic-era column name `user_id`. Same is almost certainly true for `preference.user_id` and `reading_event.user_id` (will surface as soon as a signed-in user opens reading-view or saves a preference).

## Root cause

The production database was data-migrated from old Neon Postgres → Supabase Postgres in a way that preserved the alembic-era column names. The `supabase db push` step that would have aligned production with `supabase/migrations/*.sql` was either never run or applied a different schema. Local Supabase stacks and per-PR Supabase branches both have `owner_id` (because the initial_schema migration creates it that way), which is why CI a11y is green — but production is in a divergent state.

## Effort + priority

- Effort: S (one new migration + one operator `supabase db push`)
- Priority: P0 — same as #103. Live site is broken.

## Acceptance criteria

- New Supabase migration file that idempotently renames `user_id → owner_id` on `passage`, `preference`, `reading_event` (including any indexes).
- Production database has `owner_id` on all three tables (verified by operator-run query).
- Paste + reading view + preference toggle all work end-to-end on production.
- The migration is a no-op on databases that already have `owner_id` (local + per-PR Supabase branches), so re-running it is safe.

cubrox:
## Summary

After #104 merged, the live site moved past the type-coercion crash and hit the NEXT layer's failure: production's `passage` table doesn't have an `owner_id` column — it still carries the alembic-era `user_id` name. The Supabase migrations in this repo were written with `owner_id` (post-SUPA-2c shape), but production's database was never realigned. Same is almost certainly true for `preference.user_id` and `reading_event.user_id`.

Closes #105.

## Root cause

```
psycopg.errors.UndefinedColumn: column "owner_id" of relation "passage" does not exist
```

The production Supabase Postgres was data-migrated from the old Neon Postgres in a way that preserved the alembic-era column names. The `supabase db push` step that would have applied the `owner_id`-shaped migrations to production was either never run or applied a different schema. Local Supabase stacks and per-PR Supabase branches both have `owner_id` (because `supabase/migrations/20260524015233_initial_schema.sql` creates it that way), which is why CI a11y is green — but production is in a divergent state.

## What's in this PR

One new migration file: `supabase/migrations/20260526165646_rename_user_id_to_owner_id.sql`. **Idempotent by design:**

- `ALTER TABLE ... RENAME COLUMN user_id TO owner_id` guarded by `IF EXISTS` on the old column. No-op on databases that already have `owner_id`.
- Index renames (`ix_passage_user_id_created_at` → `ix_passage_owner_id_created_at`, same for `reading_event`) with the same guard pattern.
- FK constraint cleanup: drops any existing FK on `owner_id` (which may have pointed at the deleted `public.user` table from the alembic era) and adds a fresh FK against `auth.users(id)` on all three tables. The `DROP CONSTRAINT IF EXISTS` makes the `ADD CONSTRAINT` step idempotent across all current database states.

## Why idempotent matters here

| Database state | What this migration does |
|---|---|
| Production (today): user_id columns | Renames to owner_id, fixes the bug |
| Local Supabase: already owner_id | No-op (guards skip the renames) |
| Per-PR Supabase branches: already owner_id | No-op |
| Re-run on prod after success | No-op (the guards make it a no-op once owner_id exists) |

## What Laura needs to do after merge

The Supabase GitHub App auto-applies migrations to per-PR branches, but **NOT to the production project**. After merging:

```bash
# from the repo root, with supabase CLI installed:
supabase link --project-ref gnswmcgaztcxslirulwm  # if not already linked
supabase db push --linked
```

Then verify in the Supabase Dashboard SQL Editor:

```sql
SELECT table_name, column_name
FROM information_schema.columns
WHERE table_schema = 'public'
  AND table_name IN ('passage', 'preference', 'reading_event')
  AND column_name IN ('user_id', 'owner_id');
-- Expect: three `owner_id` rows, zero `user_id` rows.
```

After that, paste + upload + reading view + preference toggle should all work end-to-end. The previous fix (#104) handled the Python-side type contract; this PR handles the database-side column name.

## Test plan

- [x] `uv run pytest` — 214 passed (no new tests; the schema change is exercised by the existing test suite via the local Supabase stack in CI a11y)
- [x] `uv run ruff check .` — clean
- [x] Pre-push hook passed
- [ ] CI green on this PR (the a11y job will run this migration on its local Supabase stack — that's the syntax+semantics test for the SQL)
- [ ] After merge: operator runs `supabase db push --linked` and confirms the verification query above
- [ ] Manual: paste a passage on production → land on `/read/{id}`. Same for PDF upload.

## Why this couldn't be tested with new pytest cases

Pytest uses in-memory SQLite via conftest. The SQL migration only runs against Postgres (Supabase's local stack or the production project). The CI a11y job is what exercises the migration end-to-end — if the SQL had a syntax error or referenced a non-existent column on the post-rename state, that job would fail. Open to adding a Postgres-backed test if you'd like more coverage; out of scope for the P0 fix today.

## Related

- Depends on #104 (the type-coercion fix; that change is what made this bug surface — before #104, requests crashed at the SQLAlchemy `.hex` layer and never reached the INSERT)
- This is the second layer of the same end-to-end bug (paste/upload 500). After this PR + the `supabase db push`, the live ingestion flow should work.

### DIFF
--- a/supabase/migrations/20260526165646_rename_user_id_to_owner_id.sql
+++ b/supabase/migrations/20260526165646_rename_user_id_to_owner_id.sql
@@ -1,0 +1,129 @@
+-- Idempotent rename of `user_id` → `owner_id` on tenant-scoped tables (BUG-4 / #105).
+--
+-- Background: the production Supabase Postgres was data-migrated from
+-- the old alembic-era Neon database, which used the column name
+-- `user_id`. The SUPA-2c rename (`user_id` → `owner_id`) was applied to
+-- the SQLModel definitions + the supabase/migrations/*.sql files
+-- (which use `owner_id` from their initial state), but the production
+-- database itself was never realigned. Surface symptom: paste fails
+-- with `psycopg.errors.UndefinedColumn: column "owner_id" of relation
+-- "passage" does not exist` after the #104 type-coercion fix moved
+-- past the previous crash. See #105.
+--
+-- This migration is IDEMPOTENT:
+--   - On production (still has `user_id`): does the rename.
+--   - On local Supabase / per-PR branches (already `owner_id`): no-op.
+--   - Re-running is safe.
+--
+-- Why IF EXISTS in every branch: the initial_schema migration already
+-- creates the column as `owner_id` for fresh databases, so the IF
+-- EXISTS guards make this migration land cleanly regardless of whether
+-- a given database was bootstrapped from the alembic-era schema or
+-- from a fresh `supabase db push`.
+
+DO $$
+BEGIN
+    -- passage.user_id → owner_id
+    IF EXISTS (
+        SELECT 1 FROM information_schema.columns
+        WHERE table_schema = 'public'
+          AND table_name   = 'passage'
+          AND column_name  = 'user_id'
+    ) THEN
+        ALTER TABLE public.passage RENAME COLUMN user_id TO owner_id;
+    END IF;
+
+    -- preference.user_id → owner_id  (also the primary key — the rename
+    -- automatically renames the implicit PK constraint).
+    IF EXISTS (
+        SELECT 1 FROM information_schema.columns
+        WHERE table_schema = 'public'
+          AND table_name   = 'preference'
+          AND column_name  = 'user_id'
+    ) THEN
+        ALTER TABLE public.preference RENAME COLUMN user_id TO owner_id;
+    END IF;
+
+    -- reading_event.user_id → owner_id
+    IF EXISTS (
+        SELECT 1 FROM information_schema.columns
+        WHERE table_schema = 'public'
+          AND table_name   = 'reading_event'
+          AND column_name  = 'user_id'
+    ) THEN
+        ALTER TABLE public.reading_event RENAME COLUMN user_id TO owner_id;
+    END IF;
+END $$;
+
+-- Rename indexes that reference the old column name.
+-- (Postgres preserves the index logically when a column is renamed,
+-- but the INDEX NAME stays as it was, so we rename for grep-ability.
+-- The composite indexes were defined in the alembic-era migrations as
+-- `ix_<table>_user_id_created_at`; rename to match the supabase-era
+-- `ix_<table>_owner_id_created_at` shape.)
+DO $$
+BEGIN
+    IF EXISTS (
+        SELECT 1 FROM pg_indexes
+        WHERE schemaname = 'public'
+          AND indexname  = 'ix_passage_user_id_created_at'
+    ) THEN
+        ALTER INDEX public.ix_passage_user_id_created_at
+            RENAME TO ix_passage_owner_id_created_at;
+    END IF;
+
+    IF EXISTS (
+        SELECT 1 FROM pg_indexes
+        WHERE schemaname = 'public'
+          AND indexname  = 'ix_reading_event_user_id_created_at'
+    ) THEN
+        ALTER INDEX public.ix_reading_event_user_id_created_at
+            RENAME TO ix_reading_event_owner_id_created_at;
+    END IF;
+END $$;
+
+-- Drop any leftover alembic-era FK constraints that pointed at the
+-- deleted `public.user` table. Recreate the FK against `auth.users`
+-- to match what the supabase-era initial_schema.sql declares.
+--
+-- Why this matters: if the data-migration preserved the column but the
+-- old `public.user` table was dropped on the way to Supabase, the FK
+-- constraint is dangling. On a database where the FK already points at
+-- auth.users (fresh supabase bootstrap), the DROP IF EXISTS is a no-op
+-- and the ADD CONSTRAINT either succeeds (if missing) or duplicates —
+-- so we DROP IF EXISTS first to make ADD idempotent.
+DO $$
+DECLARE
+    fk_record RECORD;
+BEGIN
+    -- Drop ANY foreign-key constraint on passage.owner_id /
+    -- preference.owner_id / reading_event.owner_id, regardless of
+    -- target. We re-add the auth.users FK below.
+    FOR fk_record IN
+        SELECT conname, conrelid::regclass AS table_name
+        FROM pg_constraint
+        WHERE contype = 'f'
+          AND conrelid IN (
+              'public.passage'::regclass,
+              'public.preference'::regclass,
+              'public.reading_event'::regclass
+          )
+    LOOP
+        EXECUTE format(
+            'ALTER TABLE %s DROP CONSTRAINT IF EXISTS %I',
+            fk_record.table_name, fk_record.conname
+        );
+    END LOOP;
+END $$;
+
+ALTER TABLE public.passage
+    ADD CONSTRAINT passage_owner_id_fkey
+        FOREIGN KEY (owner_id) REFERENCES auth.users(id) ON DELETE CASCADE;
+
+ALTER TABLE public.preference
+    ADD CONSTRAINT preference_owner_id_fkey
+        FOREIGN KEY (owner_id) REFERENCES auth.users(id) ON DELETE CASCADE;
+
+ALTER TABLE public.reading_event
+    ADD CONSTRAINT reading_event_owner_id_fkey
+        FOREIGN KEY (owner_id) REFERENCES auth.users(id) ON DELETE CASCADE;
--- a/supabase/migrations/20260526165646_rename_user_id_to_owner_id.sql
+++ b/supabase/migrations/20260526165646_rename_user_id_to_owner_id.sql
@@ -82,32 +82,55 @@ BEGIN
     END IF;
 END $$;
 
--- Drop any leftover alembic-era FK constraints that pointed at the
--- deleted `public.user` table. Recreate the FK against `auth.users`
--- to match what the supabase-era initial_schema.sql declares.
+-- FK constraint cleanup + recreate.
 --
--- Why this matters: if the data-migration preserved the column but the
--- old `public.user` table was dropped on the way to Supabase, the FK
--- constraint is dangling. On a database where the FK already points at
--- auth.users (fresh supabase bootstrap), the DROP IF EXISTS is a no-op
--- and the ADD CONSTRAINT either succeeds (if missing) or duplicates —
--- so we DROP IF EXISTS first to make ADD idempotent.
+-- We need to handle three FK relationships across these three tables:
+--   * passage.owner_id        → auth.users(id)
+--   * preference.owner_id     → auth.users(id)
+--   * reading_event.owner_id  → auth.users(id)
+--   * reading_event.passage_id → public.passage(id)   (NOT touched by the rename)
+--
+-- The alembic-era FKs pointed at `public.user(id)` (now-deleted) for
+-- the owner side. The supabase-era FKs point at `auth.users(id)`. So
+-- we drop ONLY the owner-side FKs (whether they reference public.user
+-- or auth.users) and re-add them against auth.users. The passage_id
+-- FK is left alone — its target (public.passage) exists in both eras
+-- and the rename doesn't touch it.
+--
+-- The DROP loop is narrowed to FKs whose target is `auth.users` or
+-- `public.user` to avoid silently dropping unrelated FKs (most
+-- importantly `reading_event.passage_id_fkey`, which the supabase-era
+-- schema needs preserved). See #106 review.
+--
+-- We use `NOT VALID` on the auth.users FKs and a follow-up
+-- `VALIDATE CONSTRAINT` step. This is the standard Postgres pattern
+-- for adding an FK to a table that may have rows whose foreign-key
+-- values predate the constraint (e.g., orphan owner_id UUIDs left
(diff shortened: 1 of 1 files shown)


cubrox:
## PR Review — #106

> ⚠️ **Conflict of interest disclosure:** This PR was authored by the same Claude session running this review. Per `feedback_review_consistency` lesson, I'm being extra skeptical about claims that the same author would not have caught themselves. Treat this review as authored by the same brain that wrote the migration — Laura should weigh the findings accordingly.

---

### Context

- Linked issue: #105 (BUG-4, P0). PR body says `Closes #105` and GitHub confirms the auto-close link is wired.
- All 21 CI checks green (a11y at 2m28s, Supabase Preview green at the per-PR branch).
- Production log dump shows the empirical state: `psycopg.errors.UndefinedColumn: column "owner_id" of relation "passage" does not exist` from `agile-flow-app-00104-6vn` at `2026-05-26T16:49:36Z` — confirms prod is in alembic-era shape.

---

### Required Changes (blocking)

**1. FK-drop loop silently drops `reading_event_passage_id_fkey` and never recreates it.**

The loop at line ~95-117 selects `WHERE contype = 'f' AND conrelid IN (passage, preference, reading_event)` with no filter on `conkey` (column). On the alembic-era production schema, `reading_event` has TWO foreign keys (per `alembic/versions/007_*.py` from commit `e6be06b`):

- `reading_event.user_id → user.id` (with `ondelete='CASCADE'`)
- `reading_event.passage_id → passage.id` (with `ondelete='CASCADE'`)

The loop drops BOTH. The bottom of the migration only re-adds `reading_event_owner_id_fkey`. After this migration runs on prod, `reading_event.passage_id` has no FK constraint — referential integrity AND the cascade-delete-on-passage-delete behavior are silently lost. The supabase-era `initial_schema.sql` (line 57) defines this FK; we'd be regressing from that intended state.

This was not caught by CI a11y because the local Supabase stack starts from `initial_schema.sql`, which already has `owner_id` and ENABLES the migration's `IF EXISTS` guards to skip the rename — so the FK loop selects only the supabase-shape FKs (`*_owner_id_fkey` pointing at `auth.users`), drops them, re-adds the same set. Idempotent there. But on prod, the loop picks up the extra `passage_id` FK and we never restore it.

Fix: either (a) filter the loop by column (`AND conkey @> ARRAY[(SELECT attnum FROM pg_attribute WHERE attrelid = conrelid AND attname IN ('user_id', 'owner_id'))]::smallint[]`), or (b) drop FKs by specific name with `DROP CONSTRAINT IF EXISTS passage_user_id_fkey, passage_owner_id_fkey` etc and re-add the `reading_event_passage_id_fkey` explicitly. The second is simpler and clearer.

**2. `ADD CONSTRAINT ... REFERENCES auth.users(id)` will fail if any prod `owner_id` value has no matching `auth.users.id`.**

`ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY` validates existing rows at constraint-creation time (unless `NOT VALID` is specified). On the alembic→Supabase data migration, the `passage.user_id` UUIDs were copied from the dropped `public.user` table. Did every one of those UUIDs end up in `auth.users.id`? If the data migration didn't bridge user accounts row-for-row (or if any users were ever deleted from `auth.users` since), the `ADD CONSTRAINT` step will fail with `insert or update on table "passage" violates foreign key constraint`, aborting the entire migration in mid-transaction.

This is unknowable from the repo state — only Laura can verify by querying:

```sql
SELECT count(*) FROM public.passage
  WHERE owner_id NOT IN (SELECT id FROM auth.users);
-- repeat for preference, reading_event
```

If the count is non-zero, options are (a) clean up orphaned rows before running the migration, (b) add `NOT VALID` to the ADD CONSTRAINT clauses (skips initial validation; still enforced for new writes), or (c) accept that the migration will fail and require operator intervention. The PR currently does none of these and the failure mode is "prod still broken AND now you have a half-applied migration."

---

**1. CI gap — the green a11y run only proves no-op idempotency, not the rename branch.**

The local Supabase stack and per-PR Supabase branches both bootstrap from `initial_schema.sql`, so they hit the migration with `owner_id` already present. The `IF EXISTS` guards turn the entire migration into a no-op there. **No CI run has ever exercised the rename branch.** The first time the `ALTER TABLE ... RENAME COLUMN user_id TO owner_id` code path executes is on production. This is acceptable for a P0 hotfix but worth surfacing — the SQL is syntactically valid and reads correctly to me, but no executable test has confirmed the rename branch works end-to-end. A pgTAP test or a script that bootstraps a Postgres with the alembic schema and then applies the migration would close this gap; out of scope for the P0.

**2. Migration-history mismatch risk on `supabase db push --linked`.**

The PR body instructs Laura to run `supabase db push --linked` after merge. The Supabase CLI tracks applied migrations in `supabase_migrations.schema_migrations`. If prod's schema came from the alembic→Neon→Supabase data migration without ever populating that tracking table, the CLI will try to apply `20260524015233_initial_schema.sql` FIRST (because no entry exists for it), and that will fail with `relation "public.passage" already exists`.

The escape hatch is `supabase migration repair --status applied <version>` to mark the prior two migrations as already-applied before pushing the new one. Or, simpler: paste the new migration's SQL directly into the Supabase Dashboard SQL Editor and skip `db push` entirely (operator runs migration manually, then runs `supabase migration repair --status applied 20260526165646` to align state).

The PR body should either (a) include a pre-flight check (`SELECT version FROM supabase_migrations.schema_migrations`) and the repair path, or (b) recommend the dashboard SQL-editor route as the safer hotfix path. As written, Laura is likely to hit the "table already exists" error and need to figure out the recovery on her own under production pressure.

**3. RLS policies still missing on prod after this migration.**

If prod skipped `initial_schema.sql`, it also skipped `20260524015234_rls_policies.sql`. After this PR's migration runs, prod has the right column names but no `ENABLE ROW LEVEL SECURITY` and no policies on the three tables. The bug fix (paste/upload) doesn't depend on RLS being present — it works because the DATABASE_URL connects as `postgres` (superuser, bypasses RLS). But this means **any client that ever connects through the anon-key Data API instead of the server's superuser connection has unmediated read/write access to all users' data**. Today the app doesn't use the Data API from clients, so the exposure is theoretical, but it's a real security gap and the SUPA-2 ticket (#81) explicitly called out RLS as load-bearing. Worth a follow-up ticket to align prod RLS state with the supabase-era intent.

**4. Migration is idempotent for the column rename, but constraint-add is only idempotent because of (1).**

Tracing the migration mentally on each of three states:

| State | Behavior |
|---|---|
| A: prod (alembic shape, `user_id`, FKs to deleted `public.user`) | Rename branch runs. FK loop drops alembic FKs (including `reading_event_passage_id_fkey` — see Required #1). ADD CONSTRAINT runs (may fail — see Required #2). |
| B: per-PR Supabase branch (already `owner_id`, FKs to `auth.users`) | Rename guards skip. FK loop drops the three `*_owner_id_fkey` constraints. ADD CONSTRAINT recreates them. Net no-op. ✅ |
| C: re-run on prod after one successful application (now `owner_id`, FKs to `auth.users`, no `passage_id` FK) | Rename guards skip. FK loop drops `*_owner_id_fkey`. ADD CONSTRAINT recreates them. The missing `passage_id` FK remains missing across re-runs. ✅ idempotent re-run, but the gap from the first application persists. |

So the idempotency claim is correct as stated, but the bug from Required #1 is a one-shot regression that re-runs don't heal.

---

### Issue close hygiene

`Closes #105` directive verified — GitHub's `closedByPullRequestsReferences` confirms auto-close on merge. ✅

---

### Recommendation

---

**Result:** NO-GO
required_changes: 2
suggestions: 4
escalation: Yes — Required #1 (loop drops `reading_event_passage_id_fkey` silently) and Required #2 (FK validation may abort migration on prod if orphan rows exist) are both blocking. Required #1 is the higher-confidence find — it is structurally certain given the alembic-era schema in commit `e6be06b`. Required #2 needs a pre-flight orphan check from Laura before deciding the right escape hatch (`NOT VALID`, data cleanup, or accept-and-recover).

Recommended path: fix Required #1 by adding `reading_event_passage_id_fkey` to the ADD CONSTRAINT block (and ideally narrow the drop loop by column). For Required #2, either add `NOT VALID` to the three new constraints with a follow-up `VALIDATE CONSTRAINT` once orphans are cleaned, or have Laura run the orphan query first and only proceed if zero rows. Also update the PR body to address the migration-history mismatch (Suggestion #2) — the current "run `supabase db push --linked`" guidance will probably fail on first attempt.

The bug this PR is trying to fix is real and the migration is 80% right. The structural finding (#1) is what blocks GO under the hard rule that any Required Change = NO-GO. Iteration loop is short — one revision should clear this.

cubrox:
## PR Review — #106 (Re-review after revision 9904814)

> ⚠️ **Conflict of interest disclosure:** Same Claude session authored the fix and is now re-reviewing it. Being extra skeptical per `feedback_review_consistency`. Laura should weigh accordingly.

---

### Context

- Revision commit: `9904814` — "narrow FK drop scope + use NOT VALID (review #106)"
- All 21 CI checks green again (a11y 2m15s, Supabase Preview pass). Per prior review note this only proves the no-op branch; the rename branch still has not executed under test. That gap is unchanged by this revision and is not a blocker — it was a prior suggestion, not a Required Change.

---

### Status of prior Required Changes

**1. FK drop loop scope — ADDRESSED.**

The loop now joins `pg_class target` + `pg_namespace ns` and filters:

```sql
AND ns.nspname || '.' || target.relname IN ('auth.users', 'public.user')
```

This correctly narrows the drop to owner-side FKs only. `reading_event.passage_id_fkey` (target `public.passage`) is no longer in scope and survives the migration on the alembic-era prod database. Additionally, the migration now ends with a safety-net `DO` block that checks for ANY FK from `reading_event` whose target is `passage` and re-creates `reading_event_passage_id_fkey` (with `ON DELETE CASCADE`) if missing. That covers the edge case where a database had the previous-revision migration partially applied. Logic verified — the safety-net check uses `con.conrelid = 'public.reading_event'::regclass AND target.relname = 'passage'`, and the only FK from reading_event to passage is on `passage_id`, so the existence check is sound.

**2. NOT VALID on auth.users FK ADDs — ADDRESSED.**

All three `ADD CONSTRAINT ... REFERENCES auth.users(id) ON DELETE CASCADE` statements now end with `NOT VALID`:

- `passage_owner_id_fkey`
- `preference_owner_id_fkey`
- `reading_event_owner_id_fkey`

The migration's comment block correctly explains the implication: the constraint is recorded and enforced on future writes, but existing rows are NOT validated at migration time, so any pre-existing orphan `owner_id` UUIDs in prod won't abort the transaction. The comment also instructs the operator on the follow-up: run `ALTER TABLE ... VALIDATE CONSTRAINT ...` once orphans are confirmed clean. Standard Postgres pattern, correctly applied.

---

### Status of prior Suggestions (carry forward, none escalated to blocking)

| # | Suggestion | Status after revision |
|---|---|---|
| 1 | CI a11y green only proves no-op branch | Unchanged. Still acceptable for P0 hotfix. |
| 2 | `supabase db push --linked` may fail on schema_migrations mismatch | Unchanged. PR body still recommends `db push --linked` without mentioning the Dashboard SQL Editor fallback or `supabase migration repair`. Worth noting in operator instructions but not a code blocker. |
| 3 | RLS policies absent on prod | Unchanged. Out of scope for the P0; deserves a follow-up ticket. |
| 4 | Three-state idempotency tracing | Re-verified against revised file: State A (alembic-era prod) now drops only owner FKs, re-adds three NOT VALID FKs against auth.users, and the safety-net leaves `reading_event_passage_id_fkey` intact (the drop loop excluded it). State B (per-PR Supabase) no-ops as before. State C (post-success re-run on prod) no-ops because the rename guards skip, the drop loop re-runs against the three owner FKs (idempotent re-add), and the safety-net is satisfied. All three states converge to the intended end state. ✅ |

---

### New findings

None. The revision is tightly scoped to the two prior Required Changes and addresses each correctly. No new structural issues surfaced on re-read.

---

### Recommendation

The two blocking issues from the prior review are genuinely resolved. The four non-blocking suggestions remain non-blocking (suggestions 1, 2, 3 are environmental / operator-side; suggestion 4 is now re-verified and the idempotency claim holds).

Per the hard rule, zero Required Changes = GO.

---

**Result:** GO
required_changes: 0
suggestions: 4 (carried forward, none escalated)
escalation: No
note: Operator instructions in the PR body should ideally call out the `supabase migration repair` / Dashboard SQL Editor fallback for the `db push` step (Suggestion #2), but that's documentation polish on the merge instructions, not a code blocker.

Issue #79, opened 2026-05-23
EPIC: Supabase replatform — consolidate persistence + auth (drop Neon + Resend)


tck517:
## Standard Fields

- **Problem Statement**: We currently run two database vendors' worth of complexity (Neon for Postgres, hand-rolled magic-link via Resend for auth) when Supabase gives us both in one bill with effectively zero auth code to maintain. This epic consolidates persistence and authentication on Supabase, eliminating the Resend dependency and removing the ~500 LOC hand-rolled magic-link path documented in ADR-002.
- **Effort Estimate**: XL (epic; child tickets are S-L)
- **Priority**: P1

---

## Children

1. SUPA-1 — Provision Supabase project + local dev setup
2. SUPA-2 — DB migration: Alembic → Supabase SQL + RLS
3. SUPA-3 — Auth rewrite: replace hand-rolled magic-link with Supabase GoTrue
4. SUPA-4 — CI/infra cutover: Neon branching → Supabase branching
5. SUPA-5 — Test rework: Supabase local fixtures + auth test rewrite
6. SUPA-6 — Cleanup: drop deps, write ADR-003, update docs

## Reference implementation

`~/projects/feature-x/leadflo` — same FastAPI + Jinja2 + HTMX stack, uses Supabase Auth (GoTrue magic-link), `supabase-py` async client, Supabase CLI migrations, JWT in HttpOnly cookie. See in particular `src/api/routes/auth.py:50-176` (callback hash-extractor), `src/integrations/supabase/async_client.py:20-51` (singleton), `src/templates/auth/callback.html:22-61` (JS bridge), and `supabase/migrations/20251228160242_rls_policies.sql` (RLS pattern).

## Decision rationale

- Eliminates Resend as a vendor dependency
- Removes ~500 LOC of auth code we own (magic-link sender, session signing, token table, tests)
- Single billing relationship (Supabase) instead of two (Neon + Resend)
- Supabase Pro plan gives per-PR DB branching, replacing Neon's branching
- Pre-launch — no production users to migrate; throwaway data acceptable

## Out of scope

- Cloud Run runtime (stays)
- App table schemas other than user/auth (data shape unchanged, only the FK target changes)
- The 17 in-flight Neon Auth tickets (#61-#77) — obsoleted by this epic but not closed here

## Definition of Done (epic-level)

- All 6 child tickets closed.
- The Supabase project is the system of record for users + app data.
[...]

PR #114, opened 2026-05-28
Synthetic auth monitor for Supabase sign-in path (Refs #79)

cubrox:
## Summary

- Adds `scripts/smoke_auth.py` — an end-to-end synthetic monitor that drives the **live** app through the full authenticated happy path: `/auth/callback` → authed `/passages/new` → reload (session persists) → `/logout` (cookie cleared) → confirm session is gone.
- It mints a throwaway, email-confirmed Supabase session using the **same proven path** as the a11y test-seed route (`app/api/test_seed.py`): `admin.create_user(email_confirm=True)` + password grant. The resulting JWT is identical in shape to a real magic-link sign-in, so `/auth/callback` and `current_user` accept it. The throwaway user is deleted in a `finally` block.
- Adds `.github/workflows/synthetic-monitor.yml` — runs the monitor **hourly** (plus manual `workflow_dispatch`) against the production Cloud Run URL, sourcing the target URL and Supabase credentials from GCP at run time (no new GitHub secrets).

## Why

Closes the last engineering-side gap in the #79 DoD: *"a working preview demonstrates magic-link login via Supabase → session persists across reloads → logout clears session."* This makes that an **automated, repeating** check rather than a one-time manual click-through.

It specifically guards the regression that bit us during the migration: production silently kept serving the stale **Neon-backed** revision for ~24h after the cutover "shipped." Mock-based unit tests cannot detect that — only something hitting the real deployed URL with a real Supabase session can. Hourly scheduling means we'd catch a recurrence within the hour instead of by luck.

## Scope / honest caveats

- **Not verified end-to-end locally.** This Codespace has no Docker and no local Supabase stack, so I could not execute a real run here. Validated: `ruff check`, `ruff format --check`, AST parse, full `pytest` (214 pass), and `actionlint` on the new workflow. The two app behaviors the script asserts (`/auth/callback` → 303 `/passages/new`; unauthenticated `/passages/new` → 303 `/`) were confirmed by reading the handlers. **First real execution will be the first scheduled run or a manual `workflow_dispatch`.**
- **Operator prerequisite (one-time):** the workflow's GCP identity needs `roles/run.viewer` and `roles/secretmanager.secretAccessor` on the `supabase-url` / `supabase-anon-key` / `supabase-service-key` secrets. If that IAM isn't granted yet, the first scheduled run will fail on the secret-fetch step.
- **Deliberately out of scope:** Supabase's own email-link generation (we mint server-side; the POST `/login` → `sign_in_with_otp` path is unit-tested separately) and the browser-side JS hash-extractor in `/auth/callback` stage 1 (covered by the Playwright a11y suite). Documented in the script's module docstring.

## Test plan

- [ ] Merge, then trigger `Synthetic Auth Monitor` via **workflow_dispatch** once and confirm a green run end-to-end against production.
- [ ] Confirm the throwaway `smoke-*@smoke.cubrox.test` user is created and then deleted (no leaked users in the Supabase auth list).
- [ ] (Operator) confirm the workflow identity has the IAM bindings noted above.
- [ ] Let one scheduled (hourly) run fire and verify it's green.

Refs #79

### DIFF
--- a/.github/workflows/synthetic-monitor.yml
+++ b/.github/workflows/synthetic-monitor.yml
@@ -1,0 +1,140 @@
+name: Synthetic Auth Monitor
+
+# End-to-end auth smoke test against the LIVE production deploy. Mints a
+# throwaway Supabase session and drives the real app through
+# callback -> authed page -> reload -> logout (scripts/smoke_auth.py).
+#
+# Why scheduled (not just per-deploy): the regression this guards against
+# (#79) was production silently serving a stale, Neon-backed revision for
+# ~24h AFTER the cutover "shipped". A per-deploy check wouldn't have caught
+# drift that happens between deploys. Hourly catches it within the hour.
+#
+# Operator prerequisites (one-time):
+#   - The workflow's GCP identity needs `roles/run.viewer` (to read the
+#     service URL) and `roles/secretmanager.secretAccessor` on the
+#     supabase-url / supabase-anon-key / supabase-service-key secrets.
+#   - The Supabase project must allow email+password sign-in (Supabase
+#     default) — that's how the throwaway session is minted.
+
+on:
+  schedule:
+    # Hourly at :17 past, to avoid top-of-hour runner congestion.
+    - cron: '17 * * * *'
+  workflow_dispatch:
+    inputs:
+      base_url:
+        description: 'Override target base URL (defaults to the resolved Cloud Run URL)'
+        required: false
+        type: string
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  monitor:
+    name: Probe production sign-in
+    runs-on: ubuntu-latest
+    # Don't run on the upstream template repo — only on configured forks.
+    if: github.repository != 'vibeacademy/agile-flow-gcp'
+
+    env:
+      GCP_PROJECT_ID: ${{ secrets.GCP_PROJECT_ID }}
+      GCP_REGION: ${{ vars.GCP_REGION || 'us-central1' }}
+      SERVICE_NAME: ${{ vars.CLOUD_RUN_SERVICE || 'agile-flow-app' }}
+
+    steps:
+      - name: Check for required secrets
+        id: check_secrets
+        # GitHub Actions rejects `secrets.*` inside `if:` expressions, so we
+        # promote presence checks into outputs the downstream gates read.
+        # Same pattern as deploy.yml.
+        run: |
+          if [ -z "${{ secrets.GCP_PROJECT_ID }}" ]; then
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "auth_method=none" >> "$GITHUB_OUTPUT"
+            echo "GCP_PROJECT_ID not configured -- skipping synthetic monitor."
+          elif [ -n "${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}" ]; then
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+            echo "auth_method=wif" >> "$GITHUB_OUTPUT"
+          elif [ -n "${{ secrets.GCP_SA_KEY }}" ]; then
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+            echo "auth_method=sa-key" >> "$GITHUB_OUTPUT"
+          else
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            echo "auth_method=none" >> "$GITHUB_OUTPUT"
+            echo "Neither GCP_WORKLOAD_IDENTITY_PROVIDER nor GCP_SA_KEY configured -- skipping."
+          fi
+
+      - name: Checkout code
+        if: steps.check_secrets.outputs.skip != 'true'
+        uses: actions/checkout@v4
+
+      - name: Authenticate to Google Cloud (Workload Identity Federation)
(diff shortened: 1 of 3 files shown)


cubrox:
## PR Review — #114

CI: all green (lint, typecheck, build, python, a11y, **actionlint** on the workflow, node, PR preview deploy). Reviewed both new files in full + `app/api/test_seed.py` (mint pattern mirrored), `app/api/auth.py` / `app/main.py` (handlers asserted against), and `.github/workflows/deploy.yml` (GCP-auth pattern mirrored). Independent review by a separate agent corroborated the findings below.

### Requirements
- [x] Satisfies the #79 DoD demonstration (login → session persists across reloads → logout clears) as a live, scheduled probe, honestly scoped (skips browser hash-extractor + email-link generation, covered elsewhere).
- [x] **Fails safe**: `check_secrets` gate + per-step `if:` mirror `deploy.yml`, so unconfigured forks skip cleanly; a broken monitor yields a red scheduled run, not a prod outage.

### Code Quality
- [x] Assertions cross-check correctly against the real handlers: `/auth/callback?access_token` → 303 `/passages/new` + `sb-access-token` ([auth.py:177-185](app/api/auth.py#L177-L185)); `/logout` → 303 `/` + cookie clear ([auth.py:199-205](app/api/auth.py#L199-L205)); unauth `/passages/new` → 303 `/` ([main.py:40-42](app/main.py#L40-L42), httpx sends no `HX-Request`). No false-positive risk.
- [ ] **Cleanup leak on exception** — see Required change #1.

### Testing
- [x] Not e2e-runnable in the dev env (no Docker/Supabase) — documented, acceptable for a live-deploy probe; mint path is the proven `test_seed.py` one.
- [ ] **`_cookie_header_clears` is pure parsing logic with zero coverage** — see Required change #2.

### Security
- [x] Secret handling sound: `::add-mask::` on anon + service keys before `$GITHUB_ENV`; `set -euo pipefail` (no `set -x`); `$(...)` capture not echoed; no secret written to the step summary.
- [x] No injection: `inputs.base_url` is read as an env var by the script (not interpolated into shell); `gcloud` args quoted.
- [x] `permissions: id-token: write` suffices for WIF; `uv sync` installs `httpx`+`supabase` (both top-level deps).
- [x] `GET /passages/new` is render-only (no DB rows created by the probe).

### Recommendation
**NO-GO** — two required changes. Both are small; I'll address them on this branch now. None are security red flags and the job fails safe, but a monitor that leaks prod users and crashes (vs. clean-fails) on a transient upstream error shouldn't be the thing we rely on to catch drift.

### Required Changes
1. **Throwaway user leaks when `sign_in_with_password` raises.** In `mint_session`, the inline `delete_user` only runs when sign-in *returns* a falsy session — not when it *raises*. `supabase_auth.AuthApiError`/`AuthRetryableError` are not `httpx.HTTPError`, so a transient Supabase error propagates out of `mint_session` before `user_id` reaches `main()`, whose `finally` then sees `user_id=None` and skips deletion → leaked `@smoke.cubrox.test` user, plus a raw traceback. Hourly cadence makes this accumulate. Fix: learn `user_id` in `main()` immediately after create (before sign-in) so `finally` always cleans up, and catch the supabase error family so the monitor reports `FAIL:` cleanly.
2. **Add a unit test for `_cookie_header_clears`.** It's the only nontrivial pure logic in the script and currently untested; cover empty-value, `Max-Age=0`, 1970-expiry (cleared) and an active cookie (not cleared).

- Broaden `main()`'s except to the supabase `AuthError` family (covered by #1's fix) so upstream hiccups are noise-free `FAIL:`s, not tracebacks.
- Optional periodic sweep of `*@smoke.cubrox.test` users older than N hours as defense-in-depth.
- `SB_URL` isn't masked (it's a project URL, not a credential — low severity; left unmasked aids log debugging).

cubrox:
## PR Review (follow-up) — #114 → GO

Both required changes from the prior review are addressed in `bdd90c3`. Re-reviewed the diff.

### Required changes — resolved
1. **User-leak on sign-in error — fixed.** Minting is split into `create_throwaway_user()` + `sign_in()`, and `main()` now learns `user_id` the instant the user exists (before sign-in). Its `finally` deletes unconditionally, so a sign-in exception no longer orphans a user. `main()` also catches the broad Supabase-error case and prints `FAIL: …` instead of crashing with a traceback.
2. **Test coverage added — done.** New `tests/test_smoke_auth.py`: 7 tests for `_cookie_header_clears` (empty value / `Max-Age=0` / 1970-expiry / active cookie / other-cookie / empty list / multi-line), plus 2 regression tests that pin the fix — `delete_user` is called with the user id even when `sign_in_with_password` raises, and is NOT called when user creation fails.

### CI
Green on the re-run: lint, typecheck, build, actionlint, test, **python** (runs the new tests + coverage). The a11y gate is unrelated to this change (no template/app code touched) and was green on the prior commit.

### Recommendation
**GO** — the cleanup-leak and crash-vs-clean-fail issues are resolved and pinned by tests; secret handling and fail-safe behavior were already sound.

Standing reminders for the human merger (unchanged, non-blocking):
- The monitor still hasn't run end-to-end (no Docker/Supabase in the dev env) — first real run is a post-merge `workflow_dispatch`.
- Operator IAM prerequisite: the workflow identity needs `roles/run.viewer` + `secretmanager.secretAccessor` on the `supabase-*` secrets, or the first scheduled run fails on the secret-fetch step.

PR #115, opened 2026-05-29
Fix: bionic reading toggle applies immediately (no reload)

cubrox:
## Summary

- **Bug:** enabling/disabling the Bionic reading preference did nothing until you reloaded the page. The toggle only swapped the inline `<style>` block (`#reading-surface-style`), but bionic emphasis is server-rendered `<b>` markup inside the passage text — not a CSS variable — so the visible text never changed.
- **Fix:** the `bionic_enabled` toggle now re-renders the passage article server-side and returns it as an HTMX **out-of-band swap** alongside the `<style>` block, so the emphasis appears/disappears the instant you click. Bionic stays server-side per the READ-3 guardrail in `app/services/reading/bionic.py` (no client-side transform).

## How

- Extracted the reading surface into `templates/fragments/reading_surface.html` (with an optional `hx-swap-oob` attribute) so both the full page and the toggle response render it from one place.
- `POST /preferences/{key}` now returns `fragments/preference_update.html` = the `<style>` block + (only when `bionic_enabled` changed and the button sent its `passage_id`) the OOB-swapped article.
- Preference buttons send `passage_id` in `hx-vals`. CSS-only preferences (color/size/width/font/line-height) are unchanged — style swap only — so a large passage isn't re-sent on every color tweak.
- `passage_id` is **optional**: direct/unit callers that omit it still get a valid 200 with just the `<style>` block; an unknown or cross-user `passage_id` yields no OOB swap (no error, no existence leak).

## Not changed (per request)

The "Sepia text" vs "Dark text" option — they are genuinely different values (`#3d2914` brown vs `#1a1a1a` near-black), just visually close. Left as-is.

## Test plan

- [x] 5 new tests in `tests/test_preferences.py`: bionic-on OOB swaps the surface with `<b>` markup; bionic-off OOB swaps back to plain text; non-bionic toggle does NOT OOB-swap; bionic without `passage_id` returns style-only; cross-user `passage_id` yields no OOB swap.
- [x] Full suite green (219 passed); ruff check + format clean.
- [x] Eyeballed the real rendered response: bionic-on returns `<style …>` + `<article id="reading-surface" hx-swap-oob="true">…<b>…</b>…</article>`; size change returns the style block only.
- [ ] Manual: in a browser, toggle Bionic on the reading view and confirm the emphasis applies without a reload (could not run a browser in this Codespace — no Docker/Supabase).

### DIFF
--- a/app/api/reading.py
+++ b/app/api/reading.py
@@ -109,6 +109,7 @@ def update_preference(
     user: CurrentUser,
     session: SessionDep,
     value: Annotated[str, Form()],
+    passage_id: Annotated[uuid.UUID | None, Form()] = None,
 ) -> HTMLResponse:
     """Set ONE preference and return the swappable <style> fragment.
 
@@ -120,6 +121,15 @@ def update_preference(
     be allow-listed. This is the dependent invariant the READ-1
     reviewer flagged: without it, the `| safe` filter in the template
     would let a user inject arbitrary CSS.
+
+    Bionic is the one preference that is NOT a CSS variable: it injects
+    server-rendered `<b>` markup into the passage text. A `<style>`-only
+    swap therefore can't apply it — the surface would stay stale until a
+    full reload. So when `bionic_enabled` changes and the caller passes
+    the `passage_id`, we also re-render the article and return it as an
+    out-of-band swap. `passage_id` is optional (CSS-only toggles and the
+    unit tests don't send it); an unknown or cross-user id simply yields
+    no OOB swap, never an error or an existence leak.
     """
     if key not in PREFERENCE_OPTIONS:
         raise HTTPException(status_code=422, detail=f"Unknown preference key: {key!r}")
@@ -139,10 +149,16 @@ def update_preference(
     stored_pref = session.get(Preference, user.id)
     prefs = with_defaults(stored_pref.values if stored_pref else None)
 
+    passage: Passage | None = None
+    if key == "bionic_enabled" and passage_id is not None:
+        passage = session.exec(
+            select(Passage).where(Passage.id == passage_id, Passage.owner_id == user.id)  # type: ignore[arg-type]
+        ).first()
+
     return templates.TemplateResponse(
         request=request,
-        name="fragments/reader_style.html",
-        context={"prefs": prefs},
+        name="fragments/preference_update.html",
+        context={"prefs": prefs, "passage": passage},
     )
 
 
--- a/templates/fragments/preference_update.html
+++ b/templates/fragments/preference_update.html
@@ -1,0 +1,12 @@
+{#
+  Response fragment for POST /preferences/{key}.
+
+  Always returns the swappable <style> block (HTMX primary target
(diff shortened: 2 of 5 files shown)


cubrox:
## PR Review — #115

CI: all checks green (lint, typecheck, build, python, **a11y**, actionlint, node). Reviewed all 5 changed files + surrounding context (`bionic.py`, `options.py`, `reader_style.html`, full `reading.py`). Independent render + test-suite verification performed.

### Requirements
- [x] Fixes the reported bug: bionic emphasis is server-rendered `<b>` markup ([bionic.py:47-75](app/services/reading/bionic.py#L47-L75)), not CSS, so the old `<style>`-only swap could never apply it. The toggle now returns the re-rendered `<article id="reading-surface" hx-swap-oob="true">` alongside the `<style>` block, applying immediately.
- [x] Feature works end-to-end — verified by rendering the real route response: bionic-on emits `<b>` markup, bionic-off emits plain text.
- [x] No scope creep — Sepia/Dark text left untouched as requested.

### Code Quality
- [x] HTMX contract correct: primary target `#reading-surface-style` and OOB target `#reading-surface` are distinct ids (no ambiguity); HTMX processes the OOB element separately from the primary swap.
- [x] Shared `fragments/reading_surface.html` reused by page + OOB path; `oob` gates the attribute. Page include omits `oob`; default Jinja `Undefined` is falsy (no StrictUndefined — [templates.py:20](app/templates.py#L20)), so the page renders cleanly.
- [x] Whitespace safe: `<article>` is single-line, so the include newline lands between `</style>` and `<article>`, never inside the passage text.
- [x] Bandwidth-conscious: article re-render gated on `key == "bionic_enabled" and passage_id is not None` — CSS toggles never re-send passage text.

### Testing
- [x] Suite passes (24 preference tests; 219 full). 5 new tests are meaningful (assert real `<b>`/no-`<b>`, OOB presence/absence, style-block survival) and cover on, off, non-bionic-no-swap, missing `passage_id`, and cross-user.

### Security
- [x] Injection neutralized in BOTH branches: bionic path escapes inside `bionicize` ([bionic.py:66-73](app/services/reading/bionic.py#L66-L73)); non-bionic path is Jinja-autoescaped.
- [x] Ownership scoped: `Passage.owner_id == user.id`. Unknown/cross-user `passage_id` → `None` → no OOB swap, no 404, no 500 (no existence leak). The preference write still succeeds.
- [x] No secrets, no SQL/command injection.

### Recommendation
**GO**

Correct, minimal, secure, and well-tested. The OOB/HTMX contract, escaping in both branches, and ownership scoping were all verified by direct render and the passing suite. CI fully green including the a11y gate (the most relevant check for a reading-surface change).

- The reliance on Jinja `Undefined`-is-falsy for `oob` works but is subtle; a one-line comment in `fragments/reading_surface.html` (e.g. "oob is unset on the page include — falsy by default") would aid future readers.
- Manual confirmation in a real browser (toggle Bionic, confirm no reload) is still worth a human's 30 seconds — automated coverage proves the server contract, not the in-browser swap.

PR #116, opened 2026-05-29
Add back link from reading view to passage-input page

cubrox:
## Summary

- The reading view (`/read/{id}`) had **no navigation** — once you were reading a passage, there was no way back to the page where you add passages/PDFs (`/passages/new`). This adds a back link at the top of the reading view.
- Link: visible text **"← Add a passage"**. The arrow (`←`) is decorative (`aria-hidden="true"`), so the link's accessible name is just "Add a passage" — matching the destination page's heading for orientation.
- A small `.reading-nav` rule aligns the link to the reading surface's centered column (it lives outside `#reading-surface`, so it uses the page's default colors, not the user's chosen reader bg/fg — no contrast surprises).

## Why this shape (accessibility)

This product targets neurodivergent + screen-reader users, so: discernible text link (not an icon-only button), decorative glyph hidden from assistive tech, and a `<nav aria-label="Back">` landmark so screen-reader users can jump to it.

## Test plan

- [x] New test `test_reading_view_has_back_link_to_passage_input` asserts `href="/passages/new"` and the decorative `aria-hidden` arrow are present.
- [x] Full suite green (229 passed); ruff check + format clean.
- [x] Rendered the real reading view and confirmed the nav appears above the reading surface linking to `/passages/new`.
- [ ] Manual: click the link in a browser from a reading page and confirm it lands on the add-passage page (couldn't run a browser here — no Docker/Supabase).

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -22,6 +22,26 @@
   overflow-wrap: break-word;
 }
 
+/* Back link on the reading view — returns to /passages/new (the passage
+   input page). Aligned to the reading surface's centered column so it sits
+   above the text, not flush against the viewport edge. Lives outside
+   #reading-surface, so it uses the page's default colors (not the user's
+   chosen reader bg/fg). */
+.reading-nav {
+  max-width: var(--reader-max-width, 65ch);
+  margin: 1rem auto 0;
+}
+
+.back-link {
+  display: inline-block;
+  text-decoration: none;
+}
+
+.back-link:hover,
+.back-link:focus {
+  text-decoration: underline;
+}
+
 /* Preference-toggle sidebar — opens as a Pico <details> at the bottom
    of the reading view. Each <fieldset> is one preference key; each
    <button> is one allow-listed option. aria-pressed reflects the
--- a/templates/pages/reading.html
+++ b/templates/pages/reading.html
@@ -3,6 +3,10 @@
 {% block title %}Reading — Master Key{% endblock %}
 
 {% block content %}
+  <nav class="reading-nav" aria-label="Back">
+    <a href="/passages/new" class="back-link"><span aria-hidden="true">&larr;</span> Add a passage</a>
+  </nav>
+
   {% include "fragments/reader_style.html" %}
 
   {% include "fragments/reading_surface.html" %}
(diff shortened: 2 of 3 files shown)


cubrox:
## PR Review — #116

Reviewed all 3 changed files + context (`base.html`, `passages_new.html`, `reading_surface.html`). Independent review by a separate agent corroborated. All required checks (a11y, python, lint, test) green; branch up to date with main.

### Requirements
- [x] Solves the stranded-reader problem: the reading view now links back to `/passages/new` (the paste/PDF input page). Reader is no longer trapped.
- [x] Scope is tight — one nav element, two CSS rules, one test. No creep.

### Code Quality
- [x] Back link sits before the `reader_style`/`reading_surface` includes, so it doesn't disturb `<article id="reading-surface">` (pinned by `test_passage_renders_inside_reading_surface_article`) or the bionic OOB swap (that fires only on the fragment route).
- [x] `&larr;` is a static template entity (not user data) → safe, renders as ←.
- [x] `.reading-nav { max-width: var(--reader-max-width, 65ch); margin: 1rem auto 0; }` correctly aligns to the reading column; the `:root` custom property applies document-wide regardless of source order, and the `65ch` fallback is belt-and-braces.

### Accessibility (the key dimension for this product)
- [x] Accessible name is the visible text "Add a passage"; the arrow is `aria-hidden="true"` (decorative, not announced).
- [x] Only `<nav>` in the template tree → no duplicate-landmark / unique-name axe violation. The a11y CI scans the full `/read/{id}` page and passed.
- [x] Link lives outside `#reading-surface`, so it uses Pico's default link-on-background colors (WCAG AA), not the user's arbitrary reader bg/fg — the correct call for contrast.

### Testing
- [x] `test_reading_view_has_back_link_to_passage_input` pins `href="/passages/new"` and the exact `<span aria-hidden="true">&larr;</span> Add a passage`. The initial assertion was a loose `aria-hidden` substring (the page has aria-hidden elsewhere on the close beacon); tightened in `969ee46` so it's a real guard. Suite green (13 reading-view tests; 229 full).

### Security
- [x] No secrets, no injection, static markup.

### Recommendation
**GO**

Correctly and minimally solves the problem with screen-reader-friendly markup and no regressions. The one review finding (loose test assertion) is already addressed.

- The earlier `mergeState` showing `BLOCKED`/`UNSTABLE` during the run was just CI completing + the non-required "Supabase Preview" skipping — not a real blocker.

Issue #139, opened 2026-06-08
OPS-3 — production migrations don't apply automatically (root cause of BUG-4 + today's paste-500)


croissantfella:
## Standard Fields

- **Problem Statement**: There is no reliable mechanism that applies `supabase/migrations/*.sql` to the **production** database. `deploy.yml` has no migration step (it comments "no workflow-driven migration step needed; Supabase GitHub App handles it"), but in practice the Supabase GitHub App's auto-branching applies migrations to **preview** branches, not production — so schema changes silently fail to land in prod. This has now caused **two production outages**: BUG-4 (#105, `owner_id` rename never applied → "column owner_id does not exist") and today's paste-500 (COMP-5 `comprehension_enabled` never applied → INSERT failed). Both were "green CI, broken prod" and fixed by hand-running SQL against the prod DB. This will recur on every schema change until the gap is closed.
- **Effort Estimate**: M
- **Priority**: P1 (each occurrence is a production outage of a core flow)

---

### A. Environment Context

- Production deploy: `.github/workflows/deploy.yml` (push to main → Cloud Run). No migration step today.
- Migrations: `supabase/migrations/*.sql`. The production DB is Supabase project `gnswmcgaztcxslirulwm`. `PRODUCTION_DATABASE_URL` already exists as a GitHub secret (deploy.yml mounts it).
- **Key constraint**: the migrations are NOT all idempotent — `20260524015233_initial_schema.sql` uses bare `CREATE TABLE` (no `IF NOT EXISTS`). So a naive "re-apply everything" fails on a populated prod DB. Any fix must respect prod's applied-migration history (`supabase_migrations.schema_migrations`), e.g. via `supabase db push`, which only applies *pending* migrations.
- Precedent: BUG-4 (#105) was the same class of failure; it was patched by hand, not systemically.

### B. Guardrails

- **Run migrations BEFORE shifting traffic** to the new revision, and **fail the deploy if a migration fails** — better a blocked deploy than new code serving against an old schema (today's exact failure mode).
- Do not break preview deploys (Supabase GitHub App auto-branching for PRs stays as-is).
- First, **verify prod's migration-history state** (`SELECT * FROM supabase_migrations.schema_migrations;`) — if the history is out of sync with what's physically applied (plausible given two manual patches), reconcile it before enabling automated `db push`, or the first automated run will fail.
- Don't log `PRODUCTION_DATABASE_URL` (contains the password) — mask it.

[...]

PR #140, opened 2026-06-08
ci(ops): one-time production migration-ledger baseline workflow (OPS-3, 1/2)

croissantfella:
## Context

OPS-3 (#139), **PR 1 of 2**. Closes the gap that caused BUG-4 and the 2026-06-08 paste-500: nothing applies migrations to production.

**Why a baseline is needed first:** diagnosis showed production has **no `supabase_migrations` ledger** — it was never CLI-managed (its schema has been maintained by hand). So `supabase db push` can't just be added to the deploy: with no recorded history it would try to re-run `initial_schema`'s `CREATE TABLE` and fail. The existing migrations must first be registered as **already-applied**.

## What this PR adds

A **dormant, manually-triggered** (`workflow_dispatch`) workflow that:
- Runs `supabase migration repair --status applied <version>` for each `supabase/migrations/*.sql` — this **records** the version in the ledger **without executing** the SQL (creates the ledger if absent).
- Runs `supabase migration list` to verify every migration shows Applied.
- Pinned CLI `2.105.0` (CI currently uses `latest` — the anti-pattern the ticket flags); DB URL masked; guarded by a typed `baseline` confirmation input; idempotent (safe to re-run).

**Merging this changes nothing at deploy time** — it only runs when you dispatch it.

## How to use (after merge)

1. Actions → **Baseline Production Migrations** → Run workflow → type `baseline` → Run.
2. Check the **Verify ledger** step output: every migration (`20260524015233` … `20260604000000`) should show **Applied** on the remote.
3. Once confirmed, **PR 2** adds `supabase db push` to `deploy.yml` so future migrations auto-apply.

### If the baseline step fails on connection/pooler
`PRODUCTION_DATABASE_URL` must be a connection the CLI can run admin/DDL over (direct or session pooler, not the transaction pooler). If it errors there, we provide a direct-connection URL and re-run — that's exactly why this is staged separately from the deploy path.

Refs #139

### DIFF
--- a/.github/workflows/baseline-migrations.yml
+++ b/.github/workflows/baseline-migrations.yml
@@ -1,0 +1,92 @@
+name: Baseline Production Migrations
+
+# ONE-TIME, manually triggered (workflow_dispatch).
+#
+# Production was never managed by the Supabase CLI — it has no
+# `supabase_migrations.schema_migrations` ledger (verified for OPS-3 #139),
+# so its schema has been maintained by hand (which is how BUG-4 and the
+# 2026-06-08 paste-500 happened). This registers the existing
+# supabase/migrations/*.sql as ALREADY-APPLIED, WITHOUT running them
+# (`migration repair --status applied` records the version; it does not
+# execute the SQL). After this runs once and "Verify ledger" shows every
+# migration as Applied on the remote, the `supabase db push` step that
+# deploy.yml adds (PR 2) is safe — push will see them as applied and only
+# run FUTURE migrations.
+#
+# Safe to re-run: `migration repair` is idempotent.
+#
+# Operator note: PRODUCTION_DATABASE_URL must be a connection the Supabase
+# CLI can run DDL/admin over (the direct or session-pooler connection, not
+# the transaction pooler). If "Mark existing migrations as applied" fails
+# with a connection/pooler error, provide a direct-connection URL (see the
+# README that PR 2 adds) and re-run.
+
+on:
+  workflow_dispatch:
+    inputs:
+      confirm:
+        description: 'Type "baseline" to confirm registering prod migrations as applied'
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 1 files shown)


croissantfella (commented):
## PR Review — #140

### Requirements
- [x] First half of OPS-3: establish the migration ledger prod has never had
- [x] Correctly staged — dormant `workflow_dispatch`; merging it cannot affect a deploy
- [x] Does not run migration SQL — `repair --status applied` only records versions (the right primitive for an existing, already-migrated DB)

### Code Quality
- [x] CLI pinned to `2.105.0` (vs CI's `latest` — honors the ticket guardrail)
- [x] DB URL masked with `::add-mask::` before use; never echoed
- [x] Typed `baseline` confirmation guard prevents accidental dispatch
- [x] Idempotent (repair is safe to re-run); version parsed from the filename's leading timestamp
- [x] `inputs.confirm` passed via env (no shell-injection sink); secret-presence check matches the deploy.yml convention

### Testing
- [x] **actionlint green**; YAML structure verified (workflow_dispatch + 6 ordered steps); 19/19 core checks pass
- [x] No app/test surface — workflow-only

### Tracking Issue Hygiene
- [x] `Refs #139` (stays open through PR 2 + the baseline run)

### Security
- [x] `contents: read` only; no `issues`/`write` scopes; URL masked; runs only on dispatch with confirmation

### Recommendation
**GO**

Safe, dormant, well-guarded first step. The real validation is the **Verify ledger** output when you run it.

### After merge — the gate before PR 2
1. Run **Baseline Production Migrations** (type `baseline`).
2. Confirm **Verify ledger** shows every migration `20260524015233 … 20260604000000` as **Applied** on the remote.
3. If it errors on the connection (transaction pooler can't run admin/DDL), we add a direct-connection URL secret and re-run — then PR 2.

Only once the ledger is verified do we add `db push` to `deploy.yml`.

PR #142, opened 2026-06-09
ci(ops): apply DB migrations to production on deploy (OPS-3, 2/2)

### LINKED ISSUE #139: OPS-3 — production migrations don't apply automatically (root cause of BUG-4 + today's paste-500)

croissantfella:
## Context

**Closes #139 (OPS-3), PR 2 of 2.** This is the fix that stops production schema drift — the root cause of BUG-4 and the 2026-06-08 paste-500 (a column that never reached prod).

PR 1 added the baseline workflow; you ran it and the ledger now shows all 4 migrations Applied on the remote. So `supabase db push` is now safe — it sees the existing migrations as applied and runs only **new** ones.

## What this adds to `deploy.yml`

A migration step that runs **before** build/deploy:
```
Configure Docker → [Check secret → Install CLI → Apply migrations] → Build → Deploy → Route traffic
```
- `supabase db push` over the **session pooler** (`SUPABASE_DB_URL`, IPv4 — the direct host is IPv6-only and unreachable from runners).
- **Fails the deploy before shipping** if a migration fails — so we never again serve new code against an old schema (the paste-500 failure mode).
- Gated on `SUPABASE_DB_URL` presence (forks without it skip migrations rather than break); pinned CLI `2.105.0`; URL masked.
- Removes the now-false "no workflow-driven migration step needed" comment.

## Effect

From now on, merging a migration to `main` applies it to production automatically on deploy. No more manual SQL, no more drift. On a deploy with no new migrations, `db push` is a fast no-op.

## Verify-after-merge

This merge itself triggers a deploy → the new step runs `db push` against prod. Since the ledger is current, it should report **no pending migrations** (no-op) and the deploy proceeds normally. Worth a glance at the deploy run's "Apply database migrations" step to confirm.

Closes #139

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -78,10 +78,44 @@ jobs:
         if: steps.check_secrets.outputs.skip != 'true'
         run: gcloud auth configure-docker ${{ env.GCP_REGION }}-docker.pkg.dev --quiet
 
-      # Schema migrations now run via the Supabase GitHub App
-      # auto-branching integration (per SUPA-1 / #80). No
-      # workflow-driven migration step needed; the alembic chain was
-      # deleted in SUPA-2b / #87.
+      # ── Apply DB migrations to production (OPS-3 #139) ────────────────
+      #
+      # Until now nothing applied supabase/migrations/*.sql to prod — it was
+      # hand-maintained, which is how BUG-4 and the paste-500 (a column that
+      # never landed) happened. Production's migration ledger was baselined
+      # once via baseline-migrations.yml, so `supabase db push` now sees the
+      # existing migrations as applied and runs only NEW ones.
+      #
+      # Run BEFORE build/deploy so a failed migration fails the deploy
+      # instead of shipping code against an old schema. Over the SESSION
+      # POOLER (IPv4) — the direct host is IPv6-only and unreachable from
+      # Actions runners (see SUPABASE_DB_URL in baseline-migrations.yml).
+      - name: Check for migration DB secret
+        if: steps.check_secrets.outputs.skip != 'true'
+        id: check_migrate
+        run: |
+          if [ -z "${{ secrets.SUPABASE_DB_URL }}" ]; then
+            echo "migrate=false" >> "$GITHUB_OUTPUT"
+            echo "SUPABASE_DB_URL not set — skipping migrations. Set the session-pooler URL to enable auto-apply."
+          else
+            echo "migrate=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Install Supabase CLI (pinned)
+        if: steps.check_migrate.outputs.migrate == 'true'
+        uses: supabase/setup-cli@v1
+        with:
+          version: 2.105.0
+
+      - name: Apply database migrations
+        if: steps.check_migrate.outputs.migrate == 'true'
+        env:
+          DB_URL: ${{ secrets.SUPABASE_DB_URL }}
+        run: |
+          set -euo pipefail
+          # URL carries the DB password — keep it out of the logs.
+          echo "::add-mask::$DB_URL"
+          supabase db push --db-url "$DB_URL"
 
       - name: Build and push container
         if: steps.check_secrets.outputs.skip != 'true'


croissantfella (commented):
## PR Review — #142

### Requirements
- [x] Completes OPS-3: production migrations now apply automatically on deploy
- [x] Builds on the verified baseline (ledger shows all 4 applied) — `db push` runs only new migrations
- [x] Removes the stale "no migration step needed" comment (DoD item)

### Code Quality
- [x] Correct ordering: migrations run **before** build/deploy/traffic-shift → a failed migration fails the deploy instead of shipping against an old schema (the paste-500 failure mode)
- [x] Uses the IPv4 session pooler (`SUPABASE_DB_URL`), consistent with baseline-migrations.yml — avoids the IPv6 direct-host issue
- [x] Gated on secret presence (forks without it skip, don't break); `set -e` makes a `db push` failure fail the step → fail the deploy
- [x] Pinned CLI `2.105.0`; URL masked with `::add-mask::`

### Testing
- [x] actionlint green; YAML + step order verified; all core CI checks pass (both a11y runs green — no flake this time)
- [x] Workflow-only; no app/test surface

### Tracking Issue Hygiene
- [x] `Closes #139`; OPS-3 documents the problem + DoD; baseline (PR 1) verified before this

### Security
- [x] `id-token: write`/`contents: read` only (unchanged); DB URL masked; no new secret exposure

### Recommendation
**GO**

The second half of the OPS-3 fix, correctly sequenced after the baseline. From merge onward, schema changes reach prod automatically and a bad migration blocks the deploy rather than corrupting a release.

### Verify-after-merge
This merge triggers a deploy → watch the **"Apply database migrations"** step: with the ledger current it should report **no pending migrations** (no-op) and the deploy proceeds. That confirms the wiring end-to-end against prod.

### Noted (not blocking)
The DoD also suggested a post-deploy paste smoke-test (both incidents were "green CI, broken prod" because no smoke test does a DB write). Worth a small follow-up ticket so this class of drift is *detected*, not just prevented.

Issue #131, opened 2026-06-05
OPS-1 — harden production auth reliability: warm instance + monitor alerting


croissantfella:
## Standard Fields

- **Problem Statement**: Two production-reliability gaps surfaced by the 2026-06-05 M1 launch-readiness assessment. (1) The production Cloud Run service runs `--min-instances=0`, so it scales to zero and cold-starts; the 2026-06-05 15:23 synthetic-auth-monitor failure was a ~30s read timeout on `/auth/callback` caused by a cold start. This both degrades first-user sign-in (vs. the <100ms NFR) and makes the monitor flaky. (2) The synthetic monitor has no alerting — on failure it's just a red ✗ in Actions; nobody is notified and no issue is opened, so a real production auth outage could go unnoticed.
- **Effort Estimate**: S
- **Priority**: P1 (launch hardening — directly affects first-user auth experience + outage detection)

---

### A. Environment Context

- `.github/workflows/deploy.yml` — production Cloud Run deploy; `flags:` block has `--min-instances=0`.
- `.github/workflows/synthetic-monitor.yml` — hourly end-to-end auth probe (`scripts/smoke_auth.py`); currently no failure alerting. The repo already auto-opens `[ALERT]` issues from `verify-agent-restrictions.yml` via `actions/github-script` — mirror that pattern.
- Preview deploys (`preview-deploy.yml`) stay `--min-instances=0` — only production gets a warm instance.

### B. Guardrails

- Only production (`deploy.yml`) gets `--min-instances=1`; previews stay scale-to-zero (cost).
- The monitor alert MUST dedup — it runs hourly, so a sustained outage must not spam a new issue every run. Open one issue and comment on subsequent failures while it's open.
- Keep the monitor job failing (red ✗) on probe failure — the alert step is additive, not a mask.

### C. Definition of Done

- Production `--min-instances=1` (one warm instance — eliminates cold-start auth timeouts); comment explains the cost/reliability tradeoff.
- Synthetic monitor opens a deduped `[ALERT]` issue on failure (comments on the existing open one rather than filing duplicates); `issues: write` permission added.
- Preview deploys unchanged (still min-instances=0).

croissantfella:
Done — PR #132 merged and deployed.

- **Warm instance live**: the post-merge production deploy (run 27028937142) succeeded with `--min-instances=1`; production verified healthy immediately after (`/api/health` 200 in ~100ms, `/api/health/db` 200). No cold start to wait on.
[...]

PR #144, opened 2026-06-10
test(ops): add a DB write-path check (paste→read) to the synthetic monitor

### LINKED ISSUE #143: OPS-4 — post-deploy smoke test that exercises a DB write (paste)
## Standard Fields

- **Problem Statement**: The last two production incidents (BUG-4, and the 2026-06-08 paste-500 from `comprehension_enabled` schema drift) were both "green CI, broken prod" — they shipped undetected because nothing exercises a real database WRITE after deploy. The synthetic auth monitor only does auth (it never touches the `passage` table), and CI tests run against an ephemeral local DB, not production. OPS-3 (#139) now *prevents* migration drift; this ticket adds *detection* for the broader class of "prod is broken in a way only a real write reveals."
- **Effort Estimate**: S
- **Priority**: P2 (detection hardening; OPS-3 already prevents the specific recurrence)

---

### A. Environment Context

- Existing post-deploy/scheduled check: `.github/workflows/synthetic-monitor.yml` drives the live app through magic-link callback → authed page → reload → logout (`scripts/smoke_auth.py`), but does NOT create a passage.
- The paste path that broke: `POST /passages` → INSERT into `passage` → 303 → `GET /read/{id}`. A smoke test that pastes a passage and loads the reading view would have caught both incidents immediately.
- The synthetic monitor already mints a throwaway Supabase session and cleans up the user — the same harness can paste (and delete) a throwaway passage.

### B. Guardrails

- Must clean up after itself (delete the throwaway passage + user) — don't accumulate junk rows in prod.
- Don't trigger comprehension LLM spend: the reading-view GET lazy-loads questions; either assert only on the paste→reading-render (not the questions panel), or disable comprehension on the throwaway passage.
- Reuse the existing auth/session + cleanup machinery in `scripts/smoke_auth.py` rather than a parallel implementation.
- On failure, surface it the same way the auth monitor now does (the deduped `[ALERT]` issue from OPS-1 #131) so a real write-path break is visible, not a silent red ✗.

### C. Definition of Done

- A smoke check (extend `scripts/smoke_auth.py` / the synthetic monitor, or a post-deploy step in `deploy.yml`) that, against the live app: signs in → pastes a passage → loads `/read/{id}` and asserts it renders (HTTP 200 + expected content) → deletes the throwaway passage + user.
- Wired to run post-deploy and/or on the synthetic-monitor schedule; failures open/append the existing `[ALERT]` issue.
- Verified by a green run against production.

---

*Filed as the detection complement to OPS-3 (#139), which prevents migration drift. Both 2026-06 prod incidents were "green CI, broken prod" for lack of a real DB-write smoke test.*

croissantfella:
## Context

Closes #143 (OPS-4). The **detection** complement to OPS-3 (#139, which *prevents* migration drift). Both 2026-06 production incidents — BUG-4 and the paste-500 — were **"green CI, broken prod"** because nothing exercised a real database **write**: the synthetic monitor only did auth (never touching the `passage` table), and CI runs against an ephemeral local DB.

## What this adds

Extends `scripts/smoke_auth.py`'s authed flow with a write path, between the reload and logout steps:

```
POST /passages (paste) → 303 to /read/<id> → GET /read/<id> → 200 with the pasted text rendered
```

It **INSERTs a row and reads it back**, so a broken write path (schema drift, a migration that didn't land) fails the monitor instead of going unnoticed.

## Why it's clean

- **Cleanup is free** — the passage is owned by the throwaway user, so the existing user-delete in `finally` removes it via FK `ON DELETE CASCADE`. No extra teardown, no leaked rows.
- **No comprehension LLM spend** — a plain `httpx` client doesn't run HTMX, so the reading view's lazy-loaded questions panel never fires (guardrail honored).
- **No workflow change** — `synthetic-monitor.yml` already runs this script hourly, and failures now open the deduped `[ALERT]` issue (OPS-1 #131). So this is automatically wired to run on schedule with alerting.

## Verification

- `ruff` (check + format) + `mypy` clean; full suite 246 green (the script isn't unit-tested — it drives a live deploy — so the real validation is the next scheduled monitor run against prod).
- The new steps reuse the existing harness (same session, same cleanup) rather than a parallel implementation.

Closes #143

### DIFF
--- a/scripts/smoke_auth.py
+++ b/scripts/smoke_auth.py
@@ -24,6 +24,12 @@ that hits the *real* deployed URL with a *real* Supabase session can.
    - `GET /passages/new` (cookie attached) → expect 200 + the authed page.
    - `GET /passages/new` again → still 200, proving the session persists across
      reloads.
+   - `POST /passages` (paste text) → expect a 303 to `/read/<id>`, then
+     `GET /read/<id>` → 200 with the pasted text rendered. This is the WRITE
+     path (OPS-4 #143): it INSERTs a row and reads it back, catching the
+     "green CI, broken prod" failures that auth-only probing misses (BUG-4,
+     the paste-500). The passage is owned by the throwaway user, so it is
+     removed by the same cleanup (FK `ON DELETE CASCADE`).
    - `GET /logout` → expect a 303 to `/` that clears the cookie.
    - `GET /passages/new` (cookie now gone) → expect a 303 back to `/`, proving
      the session is actually gone.
@@ -155,7 +161,7 @@ def sign_in(anon: Client, email: str, password: str) -> tuple[str, str]:
 
 
 def run_flow(base_url: str, access_token: str, refresh_token: str) -> None:
-    """Drive the live app through callback → authed page → reload → logout.
+    """Drive the live app through callback → authed page → reload → paste → read → logout.
 
     Raises SmokeFailure on the first deviation from the expected contract.
     """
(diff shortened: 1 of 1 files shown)


croissantfella (commented):
## PR Review — #144

### Requirements
- [x] Adds the DB-write smoke check OPS-4 asks for (the detection complement to OPS-3)
- [x] Reuses the existing harness (same session + cleanup), not a parallel implementation (guardrail)
- [x] Cleans up after itself — passage owned by the throwaway user → removed via FK ON DELETE CASCADE on user delete (verified the constraint in initial_schema.sql:24)

### Code Quality
- [x] Inserts the write path in the right place (authed → reload → **paste → read** → logout), so it runs with a live session
- [x] Asserts the real contract: `POST /passages` → 303 to `/read/<id>`, then `GET /read/<id>` 200 + the pasted marker rendered (an INSERT + read-back)
- [x] Marker is a single alphanumeric token → renders verbatim under default prefs (no bionic split), so the substring assert is reliable
- [x] FAIL messages name the failure class ("the BUG-4 / paste-500 class") for fast triage

### Guardrails honored
- [x] No comprehension LLM spend — plain httpx doesn't run HTMX, so the lazy questions panel never fires (noted in a comment)
- [x] No new workflow — synthetic-monitor.yml already runs this script hourly; failures open the deduped [ALERT] issue (OPS-1)

### Testing
- [x] ruff (check + format) + mypy clean; full suite 246 green; all core CI checks pass
- [x] Script isn't unit-tested by design (drives a live deploy) — real validation is the next scheduled monitor run against prod

### Tracking Issue Hygiene
- [x] `Closes #143`

### Security
- [x] No new secrets; throwaway user + passage both cleaned up; synthetic marker is non-sensitive

### Recommendation
**GO**

Closes the "green CI, broken prod" detection gap with a minimal, self-cleaning extension to the existing monitor. Both recent incidents would now surface within the hour as an [ALERT] issue.

### Verify-after-merge
The next scheduled **Synthetic Auth Monitor** run (hourly) will exercise the new paste→read steps against prod — confirm it's green (and that no `smoke-*` users/passages leak in Supabase).

PR #155, opened 2026-06-12
fix(reading): width in em not ch (consistent across computers)

### LINKED ISSUE #154: BUG-6 — reading width inconsistent across computers (ch unit depends on rendered font)
## Problem

The reading-surface width preference ("Narrow/Standard/Wide/Extra wide") renders a **different column width on different computers at the same setting** — some narrower than others.

## Root cause

`max_width` was expressed in **`ch`** (`55ch`/`65ch`/`75ch`/`85ch`). `ch` is the width of the font's `0` glyph, so a `ch` measure resolves to a different pixel width depending on **which font actually renders**. The font options are **system-font stacks** (e.g. `system-ui` = San Francisco on macOS, Segoe UI on Windows; `Georgia,…,serif` falls back differently) with **no bundled webfonts**, so the rendered font — and thus `ch` — varies by machine.

## Fix

Switch `max_width` from `ch` to **`em`** (relative to `--reader-size`, a fixed px value). The column is then consistent across computers (no font-glyph dependency) and still scales with the size preference (which `ch` did, and `rem` would not). ~0.5em/char keeps the measures close to the old 55/65/75/85-character widths (28/33/38/43em). Legacy stored `ch` values are normalized to `em` so existing users get the fix without re-picking.

## DoD

- `max_width` options + default in `em`; `static/style.css` fallbacks updated.
- `with_defaults` normalizes legacy `ch` width → `em`.
- Tests updated + a normalization test; full suite green.

croissantfella:
## Context

Closes #154. The reading-width preference rendered a **different column width on different computers at the same setting** — some narrower than others.

## Root cause

`max_width` used **`ch`** (`55ch`…`85ch`). `ch` is the width of the font's `0` glyph, so a `ch` measure resolves to a different pixel width depending on **which font renders**. The font options are **system-font stacks with no bundled webfonts** (`system-ui` = San Francisco on macOS / Segoe UI on Windows; the serif stack falls back differently), so the rendered font — and thus `ch` — varied by machine.

## Fix

Switch `max_width` to **`em`** (relative to `--reader-size`, a fixed px). Now the column is:
- **consistent across computers** — no dependency on the rendered font's glyph width, and
- still **scales with the size preference** (which `ch` did; `rem` would not).

`~0.5em/char` keeps the measures close to the old 55/65/75/85-char widths → **28 / 33 / 38 / 43em**. On the default macOS/Georgia/18px setup the column is ~unchanged (~594px); other machines now match it instead of being narrower/wider.

## Changes

- `options.py` / `defaults.py`: `ch` values + default → `em`; sidebar labels updated.
- `static/style.css`: `--reader-max-width` fallbacks `65ch` → `33em`.
- `with_defaults` **normalizes a legacy stored `ch` width → `em`** (`65ch`→`33em`, etc.), so existing users get the fix without re-picking their width.

## Verification

- Full suite **264 green** (+6): a new `with_defaults` normalization test (legacy ch→em, em pass-through, unknown-ch fallback) + updated reading/preference tests.
- lint + format + mypy clean. No a11y impact (width is a valid CSS length; the a11y job's default variant just renders `33em`).

## Note

`em` keeps the typographic "measure scales with font size" behavior. The remaining (pre-existing) nuance — the back-link/controls containers resolve `em` against their own 16px font, so they're ~11% narrower than the surface — is identical to the old `ch` behavior (same ratio), not a regression.

Closes #154

### DIFF
--- a/app/services/reading/defaults.py
+++ b/app/services/reading/defaults.py
@@ -7,7 +7,8 @@ user before they've toggled anything. Per the architecture doc:
   - 18px base size
   - 1.6 line-height
   - white background, near-black foreground (high but not maximum contrast)
-  - 65ch max width (the classic readable measure)
+  - 33em max width (~65 characters; in `em` not `ch` so the column is the
+    same width across computers regardless of the rendered font)
   - bionic emphasis off (opt-in only; ships in READ-3 #17)
 
 The keys here are the canonical preference keys for the rest of the
@@ -25,8 +26,20 @@ DEFAULT_PREFERENCES: dict[str, Any] = {
     "line_height": "1.6",
     "bg": "#ffffff",
     "fg": "#1a1a1a",
-    "max_width": "65ch",
+    "max_width": "33em",
     "bionic_enabled": False,
+}
+
+
+# Map legacy `ch`-based max_width values (stored before the switch to `em`)
+# to their em equivalents, so existing users get the cross-machine-consistent
+# width without having to re-pick. Any other stale `ch` value falls back to
+# the default. See the width-unit fix.
+_LEGACY_CH_TO_EM: dict[str, str] = {
+    "55ch": "28em",
+    "65ch": "33em",
+    "75ch": "38em",
+    "85ch": "43em",
 }
 
 
@@ -36,9 +49,13 @@ def with_defaults(stored: dict[str, Any] | None) -> dict[str, Any]:
     Returns a new dict — never mutates either input. If `stored` is None
     or empty, the returned dict is exactly the defaults. If `stored`
     contains keys not in DEFAULT_PREFERENCES, they're carried through
-    (forward-compat for new keys added later).
+    (forward-compat for new keys added later). A legacy `ch` max_width is
+    normalized to its `em` equivalent on the way out.
     """
     merged = dict(DEFAULT_PREFERENCES)
     if stored:
         merged.update(stored)
+    width = merged.get("max_width")
+    if isinstance(width, str) and width.endswith("ch"):
+        merged["max_width"] = _LEGACY_CH_TO_EM.get(width, DEFAULT_PREFERENCES["max_width"])
     return merged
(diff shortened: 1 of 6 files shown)


croissantfella (commented):
## PR Review — #155

### Requirements
- [x] Fixes the reported bug: same width setting → same column width on every computer
- [x] Root cause correctly identified (`ch` = font-`0`-glyph width × system-font fallback variance)
- [x] Existing users covered — legacy stored `ch` normalized to `em`, no re-pick needed

### Code Quality
- [x] `em` is the right unit: machine-consistent (depends only on the fixed-px `--reader-size`) AND preserves the measure-scales-with-size behavior `ch` had (`rem` would lose it)
- [x] Conversion (~0.5em/char → 28/33/38/43em) keeps the default ~unchanged on the common setup; other machines converge to it
- [x] Values, default, sidebar labels, and CSS fallbacks all updated together — no orphaned `ch` left (grep clean)
- [x] `with_defaults` normalization is small, pure, and falls back to default for an unknown stale `ch`

### Testing
- [x] New `with_defaults` tests: legacy ch→em mapping, em pass-through, unknown-ch fallback, key-merge
- [x] Updated reading-view + preferences tests to the em values (incl. the "survives unrelated toggle" + sidebar-extremes assertions)
- [x] Full suite 264 green; lint + format + mypy clean; a11y unaffected (still a valid length)

### Tracking Issue Hygiene
- [x] `Closes #154`

### Security
- [x] Pure presentational change; no data/auth/secret surface

### Recommendation
**GO**

Targeted fix at the actual cause, with a graceful path for existing stored values so it "just works" after deploy.

### Verify-after-merge
On the live site, set the same width on two machines (ideally different OSes) and confirm the column matches. If you'd previously set a width, it's auto-normalized — no re-pick needed.

### Noted (not blocking)
The back-link/controls containers resolve `em` against their own 16px font (≈11% narrower than the surface) — identical ratio to the old `ch`, so not a regression. A future polish could align them precisely (e.g. compute the width server-side in px), but it's out of scope for this fix.

PR #249, opened 2026-07-01
code: return 429 with Retry-After on Supabase rate-limit for /login (closes #247)

tck517:
## Summary

Splits the catch-all `Exception → 502` in `login()` into three tiers so rate-limit errors from Supabase get their own HTTP status + user-facing message.

Root-caused during #245 postmortem 2026-07-01: the prior 502 hid a straightforward `email rate limit exceeded` (Supabase's default 30/hr shared SMTP pool) for 20 minutes because the log at Cloud Run level just said 502.

Closes #247.

## What changed

- **`app/api/auth.py`**: import `AuthApiError`; new `except AuthApiError` branch that maps rate-limit errors to 429 + `Retry-After: 300` + "Too many sign-in emails just now. Please try again in a few minutes." The non-rate-limit AuthApiError path still returns generic 502. Docstring updated.
- **`tests/test_auth_login.py`**: 3 new tests. Code-based 429, message-based 429 (belt-and-suspenders), non-rate-limit AuthApiError still 502.

## Test plan
- [x] All 274 tests pass (was 271)
- [x] Ruff check + format clean
- [x] Enumeration guard preserved (generic message for non-rate-limit path)
- [ ] CI green
- [ ] After merge: verify prod via the /tmp/otp-test.py path — Supabase rate-limit should surface as 429 not 502

## Not included (separately tracked)
- #245: bump Supabase's rate limit knob (immediate operator action)
- #246: configure custom SMTP via Resend so rate limits leave Supabase's shared pool entirely (structural fix, blocked on Resend domain-verify DNS)
- #248: update Supabase Site URL (drift)

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -29,6 +29,7 @@ from typing import Annotated
 from email_validator import EmailNotValidError, validate_email
 from fastapi import APIRouter, Depends, Form, HTTPException, Request, Response
 from fastapi.responses import HTMLResponse, RedirectResponse
+from supabase_auth.errors import AuthApiError
 
 from app.config import Settings, get_settings
 from app.integrations.supabase.auth import SUPABASE_COOKIE_NAME, current_user
@@ -81,9 +82,15 @@ def login(
     Supabase will create the user on first verify, and we deliberately
     don't surface "unknown email" so this route can't enumerate.
 
-    Only failure visible to the client is a 422 for malformed-format
-    input. Errors from Supabase (rate limit at their layer, transient)
-    surface as 502.
+    Failures visible to the client:
+      - 422 for malformed email format.
+      - 429 with Retry-After when Supabase's shared SMTP pool rate-
+        limits us (#247, from #245 postmortem) — the app maps this to
+        a user-facing "try again in a few minutes" message so HTMX/
+        browsers back off gracefully. Custom SMTP via Resend (#246)
+        moves rate limits off Supabase's shared pool structurally.
+      - 502 for anything else (network, config bug, non-rate-limit
+        Supabase error). Enumeration guard: generic message.
     """
     try:
         result = validate_email(email, check_deliverability=False)
@@ -101,9 +108,23 @@ def login(
                 "options": {"email_redirect_to": redirect_url},
             }
         )
+    except AuthApiError as exc:
+        # Supabase-side error. Distinguish rate limits from everything
+        # else so users get an actionable message and operators get a
+        # meaningful status code (429 is the standard signal that
+        # HTMX/browsers already know how to back off from). Enumeration
+        # guard: we don't echo Supabase's message verbatim for non-rate-
+        # limit errors (which can leak whether an email is known).
+        if exc.code == "over_email_send_rate_limit" or "rate limit" in exc.message.lower():
+            raise HTTPException(
+                status_code=429,
+                detail="Too many sign-in emails just now. Please try again in a few minutes.",
+                headers={"Retry-After": "300"},
+            ) from exc
+        raise HTTPException(status_code=502, detail="Sign-in unavailable") from exc
     except Exception as exc:
-        # Don't leak Supabase internals; log via Cloud Logging and
-        # return a 502 so the client knows the upstream failed.
+        # Non-Supabase failure (network, config, code bug). Same 502 as
+        # before so operators can spot the class from Cloud Logging.
         raise HTTPException(status_code=502, detail="Sign-in unavailable") from exc
 
     return GENERIC_FRAGMENT
--- a/tests/test_auth_login.py
+++ b/tests/test_auth_login.py
@@ -9,6 +9,10 @@ Covers the Definition of Done for the rewritten route (SUPA-3 / #82):
     falls back to settings.app_url otherwise (local-dev path)
   - Supabase upstream errors surface as 502 (not 500) so monitoring
     can tell server-side vs upstream failures apart
+  - Supabase rate-limit errors (429 from the shared SMTP pool) are
+    remapped to 429 with Retry-After (#247, from #245 postmortem)
+    so HTMX + browsers can back off correctly and users see an
+    actionable message instead of a generic 502
 
 Rewritten in SUPA-5 (#84). The pre-SUPA-3 file tested Resend +
 MagicLinkToken; that whole flow no longer exists.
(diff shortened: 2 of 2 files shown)


tck517:

### What I Verified

#### Technical Requirements
- [x] All CI checks green (23 checks: lint, test, python, node, a11y, build, actionlint, lint-agent-policies, json-validate, version-parity, Deploy PR Preview to Cloud Run — 1m29s, Supabase Preview)
- [x] 274 tests pass (was 271; +3 new)
- [x] Ruff check + format clean (per PR body; CI `lint` job passed in 8s)
- [x] Type-safe: `AuthApiError.code` and `.message` are the correct attribute names on the `supabase_auth.errors.AuthApiError` class (verified against upstream `supabase/auth-py` at `supabase_auth/errors.py` — AuthApiError extends AuthError with `.status`, `.message`, `.code`)
- [x] `"over_email_send_rate_limit"` is a real `ErrorCode` literal in upstream — the code-based branch will match live traffic
- [x] Follows CLAUDE.md conventions (no hardcoded URLs, no emojis in tables, conventional commit format)

#### Code Quality Assessment

**`app/api/auth.py:32,111-124`:**
- New `AuthApiError` import scoped to the errors module (correct path — `supabase_auth.errors`, not the deprecated `gotrue.errors`)
- Try/except restructured cleanly: specific-before-general (AuthApiError caught before bare Exception), so subclass matching order is right
- Uses `raise ... from exc` in both branches — preserves original traceback for Cloud Logging without swallowing signal
- Rate-limit check is belt-and-suspenders (code match OR message substring) — the code-match branch is the primary path and the message match protects against upstream changing/dropping the code field
- Docstring on `login()` explicitly enumerates all client-visible failure modes (422/429/502) with cross-references to #247, #245, #246 — helpful for future maintainers

**`tests/test_auth_login.py:85-137`:**
- Three tests cover the code-based 429 path, the message-based 429 fallback (code=None), and the non-rate-limit AuthApiError path (invalid_grant → 502)
- `AuthApiError(msg, status, code)` constructor is called with the correct positional-arg order (verified against upstream: `__init__(self, message: str, status: int, code: ErrorCode)`)
- Assertions pin the two things that actually matter: HTTP status AND `Retry-After: 300` header — not just the status
- Enumeration guard note in the test docstring is spot-on: the 429 fires before any per-email logic in Supabase

#### Feature Implementation
This PR implements #247 — Handle Supabase rate-limit errors on `/login` as 429 (not 502):
- [x] `AuthApiError` handled distinctly from generic `Exception` in `login()`
- [x] Rate-limit errors return 429 with `Retry-After: 300` and a user-friendly `detail`
- [x] New test covers rate-limit path (asserts status + Retry-After header) — actually 2 new tests for the 429 path plus 1 for the negative path
- [x] Existing tests still pass (271 → 274)
- [x] Enumeration guard preserved: non-rate-limit AuthApiError still returns generic `"Sign-in unavailable"` 502

The ticket's Happy Path step 4 ("Update HTMX login form template to display the 429 detail message") is NOT implemented. See Suggestions below — I'm not blocking on it because the DoD lists only the code/test outcomes, not the template change, and this PR does not regress the UX (see the HTMX note below).

#### Testing
- 3 new tests co-located with the existing `test_auth_login.py` suite
- Tests use the existing `supabase_mock` fixture (`side_effect = AuthApiError(...)`) — same pattern as `test_supabase_upstream_error_returns_502` immediately below them
- Edge coverage:
  - code match + rate-limit message → 429 ✓
  - code=None + rate-limit message → 429 ✓ (message-based fallback)
  - non-rate-limit code + non-rate-limit message → 502 ✓
- Not covered: code=None AND message doesn't contain "rate limit" → falls to 502. Functionally equivalent to the invalid_grant test (both hit the else branch); nice-to-have but redundant.

#### Security Assessment
- ✅ No secrets, no hardcoded URLs
- ✅ Enumeration guard preserved: the 429 detail message is invariant regardless of whether the email is registered — Supabase's project-wide rate limit fires before user lookup
- ✅ The 429 message "Too many sign-in emails just now" doesn't confirm the request was well-formed beyond "the request reached Supabase" — no incremental leak vs. the previous 502
- ✅ `raise ... from exc` preserves the traceback in Cloud Logging but the client only sees the sanitized `HTTPException.detail`

#### Documentation
- ✅ PR description explains root cause (link to #245 postmortem), the fix, and what's not included (#245/#246/#248 scoped separately)
- ✅ Docstring in `login()` updated to enumerate all failure modes
- ✅ Test module docstring updated with the new invariant

### CI/CD Status
- ✅ Continuous Integration workflow: SUCCESS (all 12 jobs green in the CI workflow)
- ✅ Preview Deploy: SUCCESS (Deploy PR Preview to Cloud Run in 1m29s — first PR in this repo to have preview deploy green since #237 provisioned fresh GCP infra)
- ✅ Supabase Preview: SUCCESS

1. **HTMX form does not display the 429 message today** — `templates/home.html` has `<form id="signin-form" hx-post="/login" hx-target="#signin-form" hx-swap="outerHTML">`. HTMX 2.x by default does NOT swap on 4xx responses, and this project has neither the `response-targets` extension nor an `hx-target-429` attribute nor a global `htmx:responseError` listener in `templates/base.html`. So today: the 429 fires, the browser silently receives the JSON body `{"detail":"..."}`, and the user sees no visible change. This is the same UX as the previous 502 — no regression — but the ticket's Happy Path step 4 explicitly called this out, and a user-facing message is the whole reason we chose 429 over "log at 502 and swallow." Recommend a follow-up ticket to either (a) add the `htmx-ext="response-targets"` script + `hx-target-4*="#signin-form"` attribute, or (b) add a small `htmx:responseError` listener that reads `event.detail.xhr.responseText`, parses the `detail` field, and swaps it into an error region. Either is <20 lines.

2. **`Retry-After: 300` is optimistic** — Supabase's default email rate-limit window is 1 hour (3600s). A client that respects `Retry-After` and retries at 5 min will still 429 for another 55 min. Two options: (a) leave 300 as an operator hint (the assumption being that #245 will bump the knob or #246 will move traffic off the shared pool by the time users encounter this), (b) bump to 1800s (30 min) as a middle ground. Not a blocker — the header is a hint, not a contract — but worth revisiting after the #245/#246 outcomes are known.

3. **The message-based fallback is broad but safe** — `"rate limit" in exc.message.lower()` will also catch `over_request_rate_limit` and `over_sms_send_rate_limit` (both of which are legit rate-limit conditions where 429 IS the right answer), so the broadness is a feature not a bug. Still, if you want stricter matching later, `exc.code in {"over_email_send_rate_limit", "over_request_rate_limit", "over_sms_send_rate_limit"} or ("rate limit" in (exc.message or "").lower())` is a tighter guard.

### Strengths

- **Root cause traceability**: PR body, code comments, tests, and docstrings all reference the #245 postmortem chain — anyone touching this in six months has a clear paper trail.
- **Belt-and-suspenders check**: the code match is the primary path but the message-substring fallback protects against upstream code changes. Small, defensive, and exactly the right instinct for a boundary with an external service.
- **Test symmetry**: the three new tests mirror each other structurally (mock, POST, assert) and read cleanly — future maintainers can add a fourth branch by copy-paste-tweak.
- **Scope discipline**: the PR body explicitly calls out that #245 (knob), #246 (Resend), and #248 (Site URL) are separately tracked. No scope creep; the code change is minimal.

### Files Changed Analysis

1. **`app/api/auth.py`** (+26/-5)
   - Adds `AuthApiError` import
   - Splits the previous single `except Exception` into `except AuthApiError` (with rate-limit branch) and `except Exception` (generic fallback)
   - Docstring expanded to enumerate all client-visible failure modes

2. **`tests/test_auth_login.py`** (+58/-0)
   - 3 new tests + docstring update
   - Uses the existing `supabase_mock` fixture pattern (`.side_effect = AuthApiError(...)`)

- Correctly maps Supabase's rate-limit errors to 429 with the right attribute access (verified against upstream `supabase_auth.errors.AuthApiError`)
- Preserves the enumeration guard for non-rate-limit errors
- Ships focused tests (3) covering the code-match, message-match, and non-rate-limit paths
- Improves operator observability (429 in Cloud Logging is signal, 502 was noise)
- Passes all 23 CI checks including preview deploy

The HTMX template not being updated is a scope-relaxation from the ticket's Happy Path but does not regress UX vs. `main` today. Recommend filing a small follow-up ticket for the HTMX response-targets integration so the user-visible message actually reaches the DOM — that would fully realize the "actionable message to users" outcome from #247's problem statement.

---

tck517:
Follow-up tickets filed from review suggestions:
- **#250** — HTMX login form should display 429 error on rate limit (P2, code)
- **#251** — Tune Retry-After header on /login 429 (P2, chore)
- **#252** — Tighten 429 detection to code whitelist (P2, chore)

Merging.

PR #271, opened 2026-07-03
chore: tune Retry-After on /login 429 from 300 to 900 seconds

tck517:
## Summary

Implements #251 during drain-run \`drain-2026-07-03T1840Z\` (anchor #269, cycle 2). Increases the \`/login\` 429 Retry-After from 300s (5 min) to 900s (15 min).

## Investigation

Ticket asked whether \`AuthApiError\` exposes Supabase's own Retry-After (option 1). **No** — verified in \`supabase_auth.errors.AuthApiError\`: only \`message\`, \`status\`, and \`code\` are surfaced. Dynamic-from-response is not available.

## Rationale for 900s

- Supabase's shared-SMTP-pool default is 30 emails per project per hour → sliding window is ~1 hour
- 300s (from #247) was optimistic — clients hitting 429 would often retry back into the same window
- 3600s (full window) would be technically accurate but feels punitive if the user has a different email in mind
- 900s is the ticket-recommended compromise

Also added a code comment explaining the choice + why dynamic-from-response isn't possible.

## Structural fix reference

#246 (custom SMTP via Resend) is the real fix — it moves rate limiting off Supabase's shared pool entirely. This ticket is just tuning the interim behavior until #246 lands.

## Change

- \`app/api/auth.py\` — Retry-After 300 → 900 + rationale comment
- \`tests/test_auth_login.py\` — assertion updated

## DoD

- [x] Retry-After reflects a value calibrated to Supabase's actual rate-limit window (or explains why dynamic isn't possible)
- [x] Test updated to match
- [x] Pre-push hook green (ruff + 274 pytest)

## Test plan

- [x] Local: \`pytest tests/test_auth_login.py\` — 9 passed
- [x] Pre-push: ruff clean, 274 pytest passing
- [ ] CI: all jobs pass
- [ ] Gate: safety:reversible — starts drain's rate-limit clock

Closes #251
Drain-run: drain-2026-07-03T1840Z (anchor #269, cycle 2/3)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #272, opened 2026-07-03
chore: tighten 429 detection to code whitelist, drop message fallback

### LINKED ISSUE #252: chore: Tighten 429 detection to code whitelist (follow-up to #247)
## Problem Statement

PR #249's fallback for 429 detection uses `"rate limit" in exc.message.lower()`, in addition to checking `exc.code == "over_email_send_rate_limit"`. Reviewer flagged this substring match as broad — a hypothetical future non-rate-limit Supabase error containing "rate limit" in its message text (e.g., "IP-level rate limit reached, check dashboard") could be misclassified as 429.

Reviewer classified this as safe (rare failure mode, and 429 is a soft error — worst case is a user seeing "try again in a few minutes" for a different error that would also benefit from waiting). Not a bug, but a mild code-smell worth addressing when convenient.

Options:
1. Drop the message fallback entirely, rely only on `exc.code`. Trade-off: if Supabase renames codes upstream, we regress silently.
2. Tighten the substring to something less generic, e.g., `"email rate limit exceeded" == exc.message.lower()` (exact match).
3. Match `exc.code` against a whitelist that includes any future rate-limit-related codes: `over_email_send_rate_limit`, `over_sms_send_rate_limit`, `over_request_rate_limit` (all real Supabase codes).

**Effort:** XS
**Priority:** P3

## A. Environment Context

- File: `app/api/auth.py:112` — the `if exc.code == "over_email_send_rate_limit" or "rate limit" in exc.message.lower():` check
- Existing tests: `tests/test_auth_login.py` — code-match test (`test_supabase_rate_limit_returns_429_with_retry_after`) + message-match test (`test_supabase_rate_limit_by_message_returns_429`). If we change the fallback, at least one of these test names may need adjustment.

## B. Guardrails

- Do NOT weaken the actual 429 detection — Supabase-py MAY not always populate `.code`, so some fallback is prudent.
- Do NOT expand the whitelist to codes that aren't genuinely rate limits (e.g., `invalid_grant`).

## C. Happy Path

1. Pick approach 3 (code whitelist) — safest and most explicit.
2. Replace the condition with `if exc.code in {"over_email_send_rate_limit", "over_sms_send_rate_limit", "over_request_rate_limit"}:` and drop the message fallback.
3. Update `test_supabase_rate_limit_by_message_returns_429` to instead test each code in the whitelist. Rename appropriately.
4. Local pytest green.

## D. Definition of Done

- `if exc.code in {...whitelist...}:` used, no substring match on the message.
- Test covers each code in the whitelist.

## Dependencies

**Blocked by:** none.

## Plan reference

Follow-up from PR #249 review (reviewer suggestion 3 of 3).

## Verification

Test suite.

tck517:
## Summary

Implements #252 during drain-run \`drain-2026-07-03T1840Z\` (anchor #269, cycle 3/3, second reversible after 30-min rate-limit pause).

Replaces the substring-based fallback in \`/login\`'s 429-detection with an explicit code whitelist. Closes the last of the three review suggestions from PR #249 review.

## Change

**Before:**
\`\`\`python
if exc.code == "over_email_send_rate_limit" or "rate limit" in exc.message.lower():
\`\`\`

**After:**
\`\`\`python
RATE_LIMIT_CODES = frozenset({
    "over_email_send_rate_limit",
    "over_sms_send_rate_limit",
    "over_request_rate_limit",
})
# ...
if exc.code in RATE_LIMIT_CODES:
\`\`\`

## Rationale (per ticket recommendation — approach 3)

- The old message fallback risked misclassifying unrelated errors whose text happened to contain "rate limit" — e.g., \`"IP-level rate limit reached, check dashboard"\` from a hypothetical future Supabase error
- The whitelist inversion: **only** these codes map to 429; everything else (including future codes we haven't seen yet) → 502 until explicitly added to the whitelist
- All three codes are real Supabase codes per \`supabase-auth-py\` — email + SMS + request rate limits

## Tests

- \`test_supabase_rate_limit_by_message_returns_429\` **replaced** with parameterized \`test_supabase_rate_limit_codes_return_429\` covering all 3 codes
- **New** \`test_supabase_rate_limit_message_only_no_code_returns_502\` — regression guard: an \`AuthApiError\` with \"rate limit\" in message but a non-whitelisted code MUST return 502, not 429

Went from 9 auth-login tests to 12.

## DoD

- [x] \`if exc.code in RATE_LIMIT_CODES:\` used, no substring match on the message
- [x] Test covers each code in the whitelist (parameterized × 3)
- [x] Regression test for the dropped substring fallback
- [x] Pre-push hook green (ruff + 274 pytest, now 277 with new tests)

## Test plan

- [x] Local: \`pytest tests/test_auth_login.py\` — 12 passed
- [x] Pre-push: ruff clean, full test suite passing
- [ ] CI: all jobs pass
- [ ] Gate: safety:reversible — 30-min rate-limit window since #271 merge has elapsed

Closes #252
Drain-run: drain-2026-07-03T1840Z (anchor #269, cycle 3/3 — closes the drain if all cycles ship)

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -64,6 +64,20 @@ RATE_LIMIT_FRAGMENT = (
     "</form>"
 )
 
+# AuthApiError codes that indicate a Supabase rate limit — map to 429 not 502.
+# The substring fallback that preceded this (#252, was in #247's PR #249) risked
+# misclassifying unrelated errors whose message happened to contain "rate limit".
+# The whitelist is the explicit inversion: only these codes are 429; everything
+# else (including future codes we haven't seen) is 502 until we add it here.
+# All three are real Supabase codes per supabase-auth-py.
+RATE_LIMIT_CODES = frozenset(
+    {
+        "over_email_send_rate_limit",
+        "over_sms_send_rate_limit",
+        "over_request_rate_limit",
+    }
+)
+
 # Cookie max-age. Supabase access tokens default to 1 hour but the
 # client refreshes them automatically; the cookie's max-age is the
 # upper bound on how long the user stays signed in across reloads.
@@ -136,7 +150,7 @@ def login(
         # HTMX/browsers already know how to back off from). Enumeration
         # guard: we don't echo Supabase's message verbatim for non-rate-
         # limit errors (which can leak whether an email is known).
-        if exc.code == "over_email_send_rate_limit" or "rate limit" in exc.message.lower():
+        if exc.code in RATE_LIMIT_CODES:
             # Return HTML fragment (not JSON HTTPException) so HTMX's
             # response-targets ext (#250) can swap it into #signin-form
             # for the real user to see instead of a silent 4xx.
--- a/tests/test_auth_login.py
+++ b/tests/test_auth_login.py
@@ -20,6 +20,7 @@ MagicLinkToken; that whole flow no longer exists.
 
 from unittest.mock import MagicMock
 
+import pytest
 from fastapi.testclient import TestClient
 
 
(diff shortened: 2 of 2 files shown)


va-reviewer (approved):
## Review — APPROVED

Reviewed as part of drain-run `drain-2026-07-03T1840Z` (anchor #269, cycle 3/3, safety:reversible).

### Verification checklist

- [x] Exactly 2 files touched: `app/api/auth.py` (+15/-1) and `tests/test_auth_login.py` (+29/-6)
- [x] `RATE_LIMIT_CODES` frozenset defined at module level with all 3 Supabase codes (`over_email_send_rate_limit`, `over_sms_send_rate_limit`, `over_request_rate_limit`)
- [x] Old substring fallback (`exc.code == "over_email_send_rate_limit" or "rate limit" in exc.message.lower()`) fully replaced with `exc.code in RATE_LIMIT_CODES`
- [x] Old `test_supabase_rate_limit_by_message_returns_429` removed
- [x] New parameterized `test_supabase_rate_limit_codes_return_429` covers all 3 whitelist codes
- [x] NEW regression guard `test_supabase_rate_limit_message_only_no_code_returns_502` — asserts an `AuthApiError` with "rate limit" in the message but a non-whitelisted code (`unexpected_failure`) returns 502
- [x] `Closes #252` present in PR body
- [x] All CI green

### Notes

- Whitelist inversion is the right call: only known rate-limit codes → 429; everything else → 502 until explicitly added. Future codes fail closed (502), not open (429), which is the safer default.
- Regression test uses a realistic false-positive message (`"IP-level rate limit reached, check dashboard"`) — good adversarial scenario.
- Inline comment explains rationale + links back to #252 for future readers.

**Verdict:** APPROVED — ready to merge.

**Result:** GO
Cycle: 3/3 (drain-run `drain-2026-07-03T1840Z` complete)
Safety class: reversible
Effort: S

PR #276, opened 2026-07-20
feat(reading): add letter, word, and paragraph spacing preferences (#165)

### LINKED ISSUE #165: READ-P2-1 — Add letter-spacing, word-spacing, and paragraph-spacing reading preferences
## Problem Statement

The reading surface has no letter-spacing, word-spacing, or paragraph-spacing controls. Typography and dyslexia research consistently finds that wider letter and word spacing reduces visual crowding and improves reading fluency for readers with dyslexia and visual-processing differences. Adding these preferences follows the exact pattern of existing CSS-variable preferences (font, size, line-height) and requires no database migration — the `preference` table already stores values as JSONB.

**Effort Estimate:** S (1-4 hr)
**Priority:** P1 — high user impact, minimal effort. Phase 2 goal: expand support toolkit based on user need.

---

### A. Environment Context

- Stack: FastAPI + Jinja2 + HTMX 2.x on Python 3.12 (see `docs/TECHNICAL-ARCHITECTURE.md`)
- No new integration points — this is a pure reading-surface change
- Existing pattern to follow: `app/services/reading/options.py` (preference allow-list + labels) and `app/services/reading/defaults.py` (default values)
- Files to create or modify:
  - MODIFY `app/services/reading/defaults.py` — add `letter_spacing`, `word_spacing`, `paragraph_spacing` entries
  - MODIFY `app/services/reading/options.py` — add those three keys to `PREFERENCE_OPTIONS` and `PREFERENCE_LABELS`
  - MODIFY `templates/fragments/reader_style.html` — add `--reader-letter-spacing`, `--reader-word-spacing`, `--reader-paragraph-spacing` CSS variable injections
  - No database migration needed — preferences stored as JSONB; new keys are forward-compatible

---

### B. Guardrails

- Do NOT add any JavaScript. These are CSS-variable preferences, same as `max_width` and `font`.
- Do NOT modify auth, passage, or comprehension routes.
- All new preference values MUST be added to the `PREFERENCE_OPTIONS` allow-list in `options.py`. The route layer validates against this list before injecting into CSS; un-allow-listed values are rejected with 422.
- Use `em` not `px` or `ch` for spacing values so they scale with the `--reader-size` preference. The `ch` unit is font-dependent and renders inconsistently across machines — see the `ch-units-font-dependent` memory note.
- Performance: preference toggle must feel instant (< 100 ms perceived latency per PRD §NFR). CSS-variable swap via HTMX `outerHTML` on `#reading-surface-style` meets this without full-page reload.

---

### C. Happy Path

1. User opens `GET /read/{passage_id}`.
2. Template renders `templates/fragments/reader_style.html`, injecting `--reader-letter-spacing: {{ prefs.letter_spacing | safe }}`, `--reader-word-spacing: {{ prefs.word_spacing | safe }}`, and `--reader-paragraph-spacing: {{ prefs.paragraph_spacing | safe }}` alongside existing CSS variables.
3. The reading surface CSS applies those variables to the `article` element (or `.reading-section`).
4. User clicks "Wider" in the letter-spacing section of the preferences sidebar.
5. HTMX fires `hx-post="/preferences/letter_spacing"` with `value=0.05em`.
6. Route validates `0.05em` against `PREFERENCE_OPTIONS["letter_spacing"]` allow-list — passes.
7. `upsert_preference(owner_id=..., key="letter_spacing", value="0.05em", session=...)` writes to the `preference` table.
8. Route returns the updated `reader_style.html` fragment via `hx-target="#reading-surface-style" hx-swap="outerHTML"`.
9. Text spacing changes immediately with no page reload.

Suggested option values:
- `letter_spacing`: `["normal", "0.03em", "0.05em", "0.08em"]` (normal = no extra spacing; 0.03–0.08em covers the dyslexia-research range)
- `word_spacing`: `["normal", "0.1em", "0.2em", "0.3em"]`
- `paragraph_spacing`: mapped to the `margin-bottom` of `.reading-section` — `["0.5em", "1em", "1.5em", "2em"]`

---

### D. Definition of Done

- `GET /read/{passage_id}` renders all three new CSS variables with the user's stored preferences (or defaults when no preference row exists).
- `POST /preferences/letter_spacing` with a value from the allow-list returns 200 and the updated style fragment.
- `POST /preferences/letter_spacing` with a value NOT in the allow-list returns 422.
- Same behaviour for `word_spacing` and `paragraph_spacing`.
- New preferences persist across requests (second `GET /read/{passage_id}` still shows the updated values).
- `uv run ruff check .` returns zero errors.
- `uv run mypy app/` returns zero errors.
- `uv run pytest` passes all tests including new tests covering each new preference key.
- PR reviewer can open the reading surface in a browser and toggle each new spacing preference to confirm immediate visual feedback with no full-page reload.

## Related Issues

- Phase 2 — Additional Reading-Support Modes

croissantfella:
## Summary

Adds three new CSS-variable reading preferences — `letter_spacing`, `word_spacing`, and `paragraph_spacing` — following the exact pattern of the existing preferences (`max_width`, `font`, `line_height`). Typography research on dyslexia and visual crowding consistently finds that wider letter and word spacing reduces crowding and improves reading fluency, which is squarely the product's purpose.

## Changes

| File | Change |
|---|---|
| `app/services/reading/options.py` | Three keys added to `PREFERENCE_OPTIONS` + friendly `PREFERENCE_LABELS` |
| `app/services/reading/defaults.py` | Defaults (`normal`, `normal`, `1em`) + docstring |
| `templates/fragments/reader_style.html` | Emits the three new `--reader-*` variables |
| `static/style.css` | Consumes them on `#reading-surface` / `.reading-section` |
| `tests/test_preferences.py` | 13 new tests (parametrized across the three keys) |

**No migration needed** — preference values are JSONB. **No sidebar template change** — the sidebar is generated from `PREFERENCE_OPTIONS`, so the three keys get fieldsets, buttons, and `aria-pressed` automatically.

## Design notes

- **`em`, not `px`/`ch`** (per ticket guardrail): `em` tracks `--reader-size`, so spacing scales with the size preference and renders identically across machines. `ch` depends on the resolved system font and varies by computer — the same reasoning already documented for `max_width`.
- **`letter_spacing`/`word_spacing` default to the CSS keyword `normal`**, so the default render is byte-for-byte unchanged; widening is opt-in.
- **⚠️ `paragraph_spacing` defaults to `1em`** — unlike the other two, its allow-list has no no-op value, and `.reading-section` previously had no `margin-bottom`. So existing users will see *slightly* more space between passage sections. Flagging explicitly in case you'd prefer `0.5em`.
- **Naming caveat:** `paragraph_spacing` maps to `margin-bottom` on `.reading-section`, which spaces the *sections* a passage is split into, not paragraphs within a section. Section text is `white-space: pre-wrap`, so intra-section paragraph breaks are literal newlines with no element to target. This matches the ticket's spec; noted here since the key name is broader than the behavior.

## Verification

- [x] `uv run ruff check .` — all checks passed
- [x] `uv run ruff format --check .` — 79 files already formatted
- [x] `uv run mypy app/` — no issues in 44 source files
- [x] `uv run pytest` — **292 passed** (13 new)
- [x] Cross-checked that every `--reader-*` variable emitted by the template is consumed by `style.css` and vice versa — no typos, no dead variables (a silent no-op risk that tests wouldn't catch)
- [ ] Manual: open the reading surface and toggle each spacing control to confirm instant feedback with no full-page reload (DoD's browser check — for the reviewer)

Closes #165

### DIFF
--- a/app/services/reading/defaults.py
+++ b/app/services/reading/defaults.py
@@ -10,6 +10,8 @@ user before they've toggled anything. Per the architecture doc:
   - 33em max width (~65 characters; in `em` not `ch` so the column is the
     same width across computers regardless of the rendered font)
   - bionic emphasis off (opt-in only; ships in READ-3 #17)
+  - `normal` letter- and word-spacing (opt-in widening; READ-P2-1 #165)
+  - 1em paragraph spacing between passage sections (READ-P2-1 #165)
 
 The keys here are the canonical preference keys for the rest of the
 codebase. READ-2 (#16) will validate user input against this same set.
@@ -28,6 +30,14 @@ DEFAULT_PREFERENCES: dict[str, Any] = {
     "fg": "#1a1a1a",
     "max_width": "33em",
     "bionic_enabled": False,
+    # READ-P2-1 (#165): `normal` is the CSS keyword for both, so the default
+    # renders exactly as before this preference existed — spacing is opt-in.
+    "letter_spacing": "normal",
+    "word_spacing": "normal",
+    # 1em ≈ one blank line between sections. Unlike the two above there is no
+    # no-op value in the allow-list; sections previously had no margin-bottom,
+    # so this default does slightly increase inter-section spacing.
+    "paragraph_spacing": "1em",
 }
 
 
--- a/app/services/reading/options.py
+++ b/app/services/reading/options.py
@@ -37,6 +37,19 @@ PREFERENCE_OPTIONS: dict[str, list[Any]] = {
     # measures.
     "max_width": ["28em", "33em", "38em", "43em"],
     "bionic_enabled": [True, False],
+    # READ-P2-1 (#165): spacing controls. Typography research on dyslexia and
+    # visual crowding finds wider letter/word spacing improves reading fluency.
+    # In `em` for the same reason as max_width above — `em` tracks --reader-size
+    # (a fixed px), so spacing scales with the size preference and renders
+    # identically across machines, whereas `ch` varies with the fallback font.
+    # `normal` is the CSS keyword, so the default is a true no-op.
+    "letter_spacing": ["normal", "0.03em", "0.05em", "0.08em"],
+    "word_spacing": ["normal", "0.1em", "0.2em", "0.3em"],
+    # Applied as margin-bottom on `.reading-section`. Note this spaces the
+    # *sections* a passage is split into, not paragraphs within a section —
+    # section text is `white-space: pre-wrap`, so intra-section paragraph breaks
+    # are literal newlines in the source text with no element to target.
+    "paragraph_spacing": ["0.5em", "1em", "1.5em", "2em"],
 }
 
 # Friendly labels for sidebar UI section headings (the <legend> per
@@ -65,6 +78,18 @@ PREFERENCE_LABELS: dict[tuple[str, Any], str] = {
     ("max_width", "33em"): "Standard",
     ("max_width", "38em"): "Wide",
     ("max_width", "43em"): "Extra wide",
+    ("letter_spacing", "normal"): "Normal",
+    ("letter_spacing", "0.03em"): "Slight",
+    ("letter_spacing", "0.05em"): "Wide",
+    ("letter_spacing", "0.08em"): "Widest",
+    ("word_spacing", "normal"): "Normal",
+    ("word_spacing", "0.1em"): "Slight",
+    ("word_spacing", "0.2em"): "Wide",
+    ("word_spacing", "0.3em"): "Widest",
+    ("paragraph_spacing", "0.5em"): "Tight",
+    ("paragraph_spacing", "1em"): "Standard",
+    ("paragraph_spacing", "1.5em"): "Loose",
+    ("paragraph_spacing", "2em"): "Loosest",
 }
 
 
--- a/static/style.css
+++ b/static/style.css
@@ -27,8 +27,16 @@
 .reading-section {
   white-space: pre-wrap;
   overflow-wrap: break-word;
-  /* READ-P2-1 (#165): user-controlled spacing between passage sections. */
-  margin-bottom: var(--reader-paragraph-spacing, 1em);
+  /* READ-P2-1 (#165): user-controlled spacing between passage sections.
+     Deliberately `padding-bottom`, NOT `margin-bottom`. The next sibling is
+     always a #question-slot, and once a question loads that slot carries
+     `.inline-question { margin: 1.5rem 0 }`. Adjacent-sibling margins collapse
+     to the larger of the two, so a margin here would be swallowed whenever the
+     preference is below 1.5rem — making "Tight" and "Standard" visually inert
+     at the default reader size. Padding never collapses, so every option stays
+     distinguishable. `.reading-section` has no background or border, so this
+     renders identically to a margin otherwise. */
+  padding-bottom: var(--reader-paragraph-spacing, 1em);
 }
 
 /* Inline comprehension question shown after each passage section. */
--- a/static/style.css
+++ b/static/style.css
@@ -13,6 +13,10 @@
   background: var(--reader-bg);
   color: var(--reader-fg);
   max-width: var(--reader-max-width);
+  /* READ-P2-1 (#165): fall back to `normal` so the surface still renders
+     correctly if the style block is ever served without these variables. */
+  letter-spacing: var(--reader-letter-spacing, normal);
+  word-spacing: var(--reader-word-spacing, normal);
   margin: 2rem auto;
   padding: 2rem;
 }
@@ -23,6 +27,8 @@
 .reading-section {
   white-space: pre-wrap;
   overflow-wrap: break-word;
+  /* READ-P2-1 (#165): user-controlled spacing between passage sections. */
+  margin-bottom: var(--reader-paragraph-spacing, 1em);
 }
 
 /* Inline comprehension question shown after each passage section. */
--- a/templates/fragments/reader_style.html
+++ b/templates/fragments/reader_style.html
@@ -6,5 +6,8 @@
     --reader-bg: {{ prefs.bg | safe }};
     --reader-fg: {{ prefs.fg | safe }};
     --reader-max-width: {{ prefs.max_width | safe }};
+    --reader-letter-spacing: {{ prefs.letter_spacing | safe }};
+    --reader-word-spacing: {{ prefs.word_spacing | safe }};
+    --reader-paragraph-spacing: {{ prefs.paragraph_spacing | safe }};
   }
 </style>
--- a/tests/test_preferences.py
+++ b/tests/test_preferences.py
@@ -17,6 +17,7 @@ Covers the Definition of Done from issue #16 (READ-2):
 import hashlib
 import uuid
 
+import pytest
 from fastapi.testclient import TestClient
 from sqlmodel import Session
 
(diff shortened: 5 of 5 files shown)


croissantfella:

### What I Verified

#### Technical Requirements
- [x] All CI checks green (lint, format, mypy, pytest 292 passed, a11y, build, actionlint, json-validate, version-parity, Supabase Preview, Deploy PR Preview)
- [x] `uv run mypy app/` clean — no new type surface (both new dicts are already typed `dict[str, list[Any]]` / `dict[str, Any]`)
- [x] No linting violations
- [x] Code follows CLAUDE.md standards (no emojis in tables, conventional commit, feature branch, no hardcoded URLs)

#### Security Assessment — the READ-1 dependent invariant

This was the thing I checked first and hardest, since `reader_style.html` uses `| safe` and the new keys feed straight into it. **The invariant holds.** Traced end to end:

1. `app/api/reading.py:155` — `if key not in PREFERENCE_OPTIONS: 422`. Key gate.
2. `app/api/reading.py:159` — `if coerced is None or coerced not in PREFERENCE_OPTIONS[key]: 422`. Value gate, membership test against the literal list.
3. Only after both gates does `upsert_preference()` run, so the DB can never hold a non-allow-listed value for these keys.
4. `with_defaults()` merges stored-over-defaults; both sides are allow-listed constants.
5. Only then does the value reach `{{ prefs.letter_spacing | safe }}`.

Both gates are **generic over the key** — they read `PREFERENCE_OPTIONS` at runtime rather than switching on known keys. That is why adding three keys to the allow-list is sufficient and no route change was needed. The three new value lists are all hardcoded string literals with no user-derived content, so there is no path from user input to the `| safe` render that bypasses the membership check.

- ✅ CSS injection through the new keys is not reachable — `test_spacing_css_injection_attempt_returns_422` asserts this directly with a `normal;}body{display:none}` payload, parametrized across all three keys
- ✅ `coerce_value()` is a pass-through for non-boolean keys, so no coercion can launder a payload into an allow-listed value
- ✅ No secrets, no hardcoded URLs, no new routes, no new auth surface

#### Ticket Guardrails (§B)
- ✅ **No JavaScript added.** Diff is Python, CSS, Jinja, tests only.
- ✅ **No changes to auth, passage, or comprehension routes.** `app/api/reading.py` is untouched by this diff.
- ✅ **All new values in the `PREFERENCE_OPTIONS` allow-list**, matching the ticket's suggested value sets exactly.
- ✅ **`em`, not `px`/`ch`.** Every non-keyword value is `em`. The rationale comment in `options.py` correctly restates the `ch`-is-font-dependent reasoning from the `max_width` fix.
- ✅ **No migration** — correct, `preference.values` is JSONB and `with_defaults()` already carries unknown keys through, so old rows upgrade cleanly with no backfill.
- ✅ **Instant toggle** — the existing `outerHTML` swap on `#reading-surface-style` covers the new variables for free; no new round trips.

#### Feature Implementation — Definition of Done (§D)
- ✅ `GET /read/{id}` renders all three variables from stored prefs or defaults — `test_read_view_renders_spacing_variables_with_defaults`
- ✅ `POST /preferences/{key}` with an allow-listed value returns 200 + updated fragment — parametrized across all three keys
- ✅ Non-allow-listed value returns 422 — parametrized across all three keys
- ✅ Persistence across requests — `test_spacing_preference_persists_across_requests` does POST then a fresh `GET /read/{id}`
- ✅ ruff / mypy / pytest all clean
- ⬜ Browser confirmation on the preview deploy — see "Note on the manual DoD item" below

#### Code Quality Assessment

**`app/services/reading/options.py`:**
- ✅ Follows the existing `max_width` pattern exactly — allow-list plus `PREFERENCE_LABELS` entries, no new abstraction invented
- ✅ Friendly labels for all 12 new (key, value) pairs, so no raw CSS values leak into button text
- ✅ `KEY_LABELS` correctly left alone — the fallback yields "Letter Spacing" / "Word Spacing" / "Paragraph Spacing", which is what you want
- ✅ The comment explaining the `paragraph_spacing` → `.reading-section` mapping is exactly the kind of comment that earns its place: it documents a non-obvious semantic gap at the point a future reader will hit it

**`app/services/reading/defaults.py`:**
- ✅ Module docstring updated alongside the dict — the docstring is treated as part of the contract, not decoration
- ✅ `normal` for letter/word spacing is the right default: byte-identical render for existing users, widening strictly opt-in

**`templates/fragments/reader_style.html` / `static/style.css`:**
- ✅ CSS `var()` fallbacks (`, normal` / `, 1em`) mean the surface degrades correctly if the style block is ever served without the variables — defensive without being noisy
- ✅ `letter-spacing`/`word-spacing` on `#reading-surface` and `margin-bottom` on `.reading-section` is the correct split given `.reading-section` was already carved out so question slots escape `pre-wrap`
- ✅ The author's cross-check that every emitted `--reader-*` variable has a consumer and vice versa is the right instinct — a typo there is a silent no-op that no test in this suite would catch

**`tests/test_preferences.py`:**
- ✅ 13 tests, parametrized over a `SPACING_KEYS` table rather than copy-pasted per key — adding a fourth spacing key later is a one-line change
- ✅ Covers the security case, not just the happy path
- ✅ `test_spacing_toggle_leaves_other_spacing_keys_untouched` asserting `pref.values == {"letter_spacing": "0.08em"}` is a nice touch — it pins the merge semantics of `upsert_preference` (store only what changed, defaults fill the rest) rather than just the rendered output

### On the two items you flagged for my judgment

**1. `paragraph_spacing` defaulting to `1em` — keep it. Do not add a no-op value.**

Keep `1em`. Two reasons.

First, the blast radius is genuinely small: this is a pre-launch surface, the change is one line-height of extra space between sections, and it moves in the direction of legibility rather than against it. Every user who wants the old density can reach it in one click via "Tight" (`0.5em`) — and `0.5em` at an 18px base is 9px, which is close enough to the previous zero-margin render that nobody will file a bug over it.

Second, and more important: I'd argue *against* adding a `0` / no-op value to the allow-list. This product exists to reduce visual crowding. An allow-list entry that lets a user collapse passage sections into an undifferentiated wall of text works directly against that purpose, and every value in an allow-list is a value you are affirmatively recommending. `0.5em` as the floor is a defensible product opinion. Ship it.

The one thing I'd push back on is if there were a meaningful population of existing users with stored preferences — there isn't, and `with_defaults()` means anyone who has already picked a `paragraph_spacing` is unaffected regardless. So the default only ever applies to users who have never expressed an opinion.

**2. The `paragraph_spacing` naming — acceptable, and if you disagree, change it in this PR rather than later.**

My honest read: the *user-facing* label is fine and arguably better than the accurate alternative. Readers do not have a mental model of "sections the passage was split into" — they see visually separated blocks of text and call them paragraphs. A sidebar legend reading "Section Spacing" would be more precise and less understandable, which is the wrong trade for an accessibility tool.

The *internal key* is the weaker half, and that is where I'd note the real consideration: the key is a JSONB key in user data. Renaming `paragraph_spacing` → `section_spacing` after merge means a backfill migration over the `preference` table, which is disproportionate for a naming nit. So this PR is the cheapest moment it will ever be. My recommendation is to keep it as-is — the comment in `options.py` documents the gap precisely and at the right location, which is the correct mitigation for a name that's 90% right — but if the naming bothers you, say so now rather than in three months.

Non-blocking either way. This is a judgment call the author correctly surfaced instead of quietly deciding.

1. **`paragraph_spacing` may be a visual no-op when comprehension questions are enabled** (`static/style.css:31`). In `reading_surface.html`, every `.reading-section` is followed by a `#question-slot-N` div. With comprehension off that slot is empty and margins collapse through it, so `margin-bottom` produces exactly the intended gap. With comprehension **on**, the slot fills with `.inline-question`, which carries `margin: 1.5rem 0` — adjacent-sibling margin collapsing means the visible gap becomes `max(paragraph_spacing, 1.5rem)`. At an 18px base, `1.5rem` is 24px, so the `0.5em` (9px) and `1em` (18px) options would produce no visible difference in that mode; only `1.5em` and `2em` would. Worth confirming visually on the preview deploy. If it holds, the fix is small (drop `.inline-question`'s vertical margin and let the section margin own the rhythm), but it's a follow-up ticket, not a merge blocker — the mapping is what the ticket specified and it behaves correctly in the comprehension-off path.

2. **Trailing margin on the last section** (`static/style.css:31`). The final `.reading-section` contributes its `margin-bottom` on top of `#reading-surface`'s `padding: 2rem`, so the bottom gutter grows as the user loosens spacing while the top stays fixed — asymmetric at `2em`. A `.reading-section:last-child { margin-bottom: 0; }` would keep the surface balanced. Cosmetic.

3. **Long-run allow-list ergonomics.** `PREFERENCE_OPTIONS` and `PREFERENCE_LABELS` are now two parallel structures that must be kept in sync by hand, and a missing label degrades silently to `str(value)` (a raw `0.08em` on a button). Not a problem at 7 keys. If Phase 2 keeps adding preferences, consider either collapsing them into one structure keyed by preference, or a test that asserts every `(key, value)` in `PREFERENCE_OPTIONS` has a `PREFERENCE_LABELS` entry. Cheap insurance against a cosmetic regression that CI currently cannot see.

### Note on the manual DoD item

The DoD asks the reviewer to open the reading surface and toggle each control in a browser. I have not done that — I reviewed statically. The preview deploy is green and the swap mechanism is unchanged from the existing preferences, so my confidence that it works is high, but **the human merger should do the browser pass**, and it is worth specifically eyeballing suggestion #1 above (toggle `paragraph_spacing` on a passage with comprehension enabled and confirm the smaller values actually move something).

### Strengths

- **Security discipline:** the author identified the `| safe` injection point as the load-bearing invariant and wrote a dedicated parametrized test with a real payload for it, rather than assuming the allow-list holds because it held last time.
- **Honest PR description:** both of the judgment calls flagged for review are real, non-obvious, and exactly the kind of thing that usually gets silently decided and discovered later. Surfacing them in the PR body is the behavior you want.
- **Pattern fidelity:** zero new abstractions. The change is legible as "three more of the thing we already do," which is why it needed no route, template, or migration changes.
- **Comments that document decisions, not code.** The `em`-vs-`ch` rationale and the `paragraph_spacing` semantic caveat both explain *why*, at the location where a future reader forms the wrong assumption.

### Files Changed Analysis

1. **`app/services/reading/options.py`** (+25) — three allow-list entries plus 12 label pairs. The security-critical file; correct.
2. **`app/services/reading/defaults.py`** (+10) — three defaults plus docstring. `normal`/`normal` are true no-ops; `1em` is the intentional behavior change discussed above.
3. **`templates/fragments/reader_style.html`** (+3) — three variable emissions, gated upstream.
4. **`static/style.css`** (+6) — two properties on `#reading-surface`, one on `.reading-section`, all with `var()` fallbacks.
5. **`tests/test_preferences.py`** (+123) — 13 parametrized tests covering render, 200, 422, injection, persistence, merge isolation, and sidebar generation.

**GO** — All quality standards met.

- The READ-1 security invariant is intact and now has dedicated regression coverage; the new keys cannot reach the `| safe` render without passing the allow-list.
- Every ticket guardrail is respected: no JS, no auth/passage/comprehension route changes, `em` units throughout.
- Every automated Definition-of-Done item is met and tested; the one remaining item is the manual browser pass, which belongs to you.
- Both flagged judgment calls are non-blocking: keep `1em`, keep the name.
- Three non-blocking suggestions, the first of which is worth a look during the browser pass.

Board note: I could not move #165 on the project board — ProjectV2 writes fail with "Resource not accessible by integration" with this token. Expected, not a blocker.

---
*Agent review completed. Human: please review my assessment and make the final merge decision. I cannot and did not approve or merge this PR.*

croissantfella:
**Addressed the margin-collapse finding** (suggestion #1) in `30b44eb` — thanks, it was a real defect, not a nit.

Verified the mechanism: `.reading-section`'s next sibling is always a `#question-slot`, and once a question loads that slot carries `.inline-question { margin: 1.5rem 0 }`. Both are adjacent siblings in a plain `<article>` (no BFC), so the margins collapse to the larger. At the default 18px reader size that meant:

| Option | Value | Was |
|---|---|---|
| Tight | 0.5em = 9px | **inert** — 1.5rem (24px) won |
| Standard | 1em = 18px | **inert** — 1.5rem (24px) won |
| Loose | 1.5em = 27px | visible |
| Loosest | 2em = 36px | visible |

Two of four options produced no visible change with comprehension enabled — which is the default (`comprehension_enabled` defaults to `True`). That also broke the ticket's own DoD ("toggle each new spacing preference to confirm immediate visual feedback"), so I treated it as in-scope for this PR rather than a follow-up.

**Fix:** `margin-bottom` → `padding-bottom`. Padding never collapses, so all four options stay distinguishable. `.reading-section` has no background or border, so it renders identically otherwise. The rationale is documented inline in `static/style.css` so nobody "tidies" it back to a margin later.

Re-ran locally: ruff clean, **292 passed**. Note this is exactly the class of bug the test suite structurally cannot catch — the tests assert the CSS *variable value*, not rendered geometry — so the manual browser pass is still worth doing, now specifically comparing Tight vs Standard **with comprehension on**.

Your other two calls (keep `1em` default, keep the `paragraph_spacing` name) — agreed, and I took the point about the JSONB key rename being now-or-expensive. Leaving both as-is.

